PR #119, opened 2023-03-18
fix: differentiate between query and path params in getSchemas

xPhentom:
While debugging, I noticed there was no difference being made between query params and path params.
So for my test with the petStore.yml I was getting true on `schemas.pathParams` even though there were no path params. It returned true because it found parameters in the query.
I'll write a test still but I just want to make sure if you also think this is indeed a bug :)

### DIFF
--- a/packages/swagger/src/generators/OperationGenerator.ts
+++ b/packages/swagger/src/generators/OperationGenerator.ts
@@ -38,6 +38,10 @@ export abstract class OperationGenerator<TOptions extends { oas: Oas } = { oas: 
       }
     })
 
+    if (!params.length) {
+      return null
+    }
+
     return params.reduce(
       (schema, pathParameters) => {
         return {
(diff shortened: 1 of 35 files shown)


packages/swagger/src/generators/OperationGenerator.ts
```diff
@@ -3,11 +3,12 @@ import { pascalCase } from 'change-case'
 import type { FileManager, File } from '@kubb/core'
 import { Generator } from '@kubb/core'
 
+import { isReference } from '../utils/isReference'
```
wouterroozeleer, line 6: Did not mean to place this import somewhere else, must have been the prettier/eslint setup

packages/swagger/src/generators/OperationGenerator.ts
```diff
@@ -53,14 +54,22 @@ export abstract class OperationGenerator<TOptions extends { oas: Oas } = { oas:
   }
 
   getSchemas(operation: Operation): OperationSchemas {
+    const parameters = operation.getParameters()
+
+    const queryParams = parameters.filter((param) => param.in === 'query')
```
wouterroozeleer: param.in could be anything but in swagger documentation it feels like these are the main/only ones?
https://swagger.io/docs/specification/describing-parameters/

stijnvanhulle:
@all-contributors please add @ xPhentom for getSchemas fix

stijnvanhulle:
https://github.com/all-contributors please add @xPhentom for getSchemas fix

stijnvanhulle:
@all-contributors  please add @xPhentom for getSchemas fix

stijnvanhulle:
@all-contributors  please add @xPhentom for code

stijnvanhulle:
@all-contributors  please add @stijnvanhulle  for code

stijnvanhulle (approved):

PR #221, opened 2023-04-28
Fix watch option

p-chan:


stijnvanhulle:
@p-chan Thanks for finding this :)

PR #436, opened 2023-09-22
fix: use of `case-anything` instead of `change-case`

stijnvanhulle:
Use of `case-anything`

- small footprint (it's [12+ times smaller](https://github.com/mesqueeb/case-anything#package-size) than the next popular case changing package!!)
- tree-shakable — only import what you need
- awesome JSDocs popup documentation on hover
- fully typed with TypeScript
- complete coverage with unit testing
- 0 dependencies


stijnvanhulle:
Case-anything has some weird behaviours that are different, closing this PR and keep `change-case`

PR #437, opened 2023-09-23
Remove encoding on regex -> Issue on zod models

meister03:
Remove Escape 
- depreciated
- regex encoding is not required --> causes issues on Zod models

```
openapi: pattern: ^[a-zA-Z0-9.-]*$
zod: new RegExp('%5E%5Ba-zA-Z0-9.-%5D*%24')
```

### DIFF
--- a/packages/swagger-zod/src/generators/ZodGenerator.ts
+++ b/packages/swagger-zod/src/generators/ZodGenerator.ts
@@ -99,7 +99,7 @@ export class ZodGenerator extends SchemaGenerator<Options, OpenAPIV3.SchemaObjec
           const isStartWithSlash = matches.startsWith('/')
           const isEndWithSlash = matches.endsWith('/')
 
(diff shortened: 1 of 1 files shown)


stijnvanhulle:
@meister03 I opened another PR related to this because removing the `escape` will result in some issues with complex Regex(also the reason why the e2e pipeline was failing).

Because our generator is text-based, we need to escape some chars to be sure they are then used correctly with Zod(`new RegExp(""`).
```
const regexp = `new RegExp('${jsStringEscape(matches.slice(isStartWithSlash ? 1 : 0, isEndWithSlash ? -1 : undefined))}')`

validationFunctions.push({ keyword: zodKeywords.matches, args: regexp })
```

https://github.com/kubb-project/kubb/pull/440 I added some examples with a new escape function `jsStringEscape`, which will handle the escaping correctly.

meister03:
Okay, let me review your pr

PR #463, opened 2023-10-09
feat(swr): shouldFetch option for swr plugin

### LINKED ISSUE #447: swagger-swr plugin doesn't support Conditional Fetching.

[swr](https://swr.vercel.app) can conditional patching by passing the first argument as `null` when calling `useSWR()', as shown in the documentation at https://swr.vercel.app/docs/conditional-fetching. This is a very useful feature that allows you to delay making an API request until the API request arguments are fetched asynchronously. However, this feature is not available using the swagger-swr plugin, which is inconvenient.

### What is the feature you are proposing to solve the problem?

I've made a modification that implements conditional patching by passing `null` as the first argument to the useSWR call when this value is `false`, given the option `shouldFetch`.

https://github.com/b6pzeusbc54tvhw5jgpyw8pwz2x6gs/kubb/commit/10f7f5d6ef5e9d84d16569162bf170c787770b99

I'll raise a PR if this direction is okay.

### What alternatives have you considered?

_No response_

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
resolve #447

### DIFF
--- a/packages/swagger-swr/src/builders/QueryBuilder.ts
+++ b/packages/swagger-swr/src/builders/QueryBuilder.ts
@@ -24,22 +24,30 @@ export class QueryBuilder extends OasBuilder<Config> {
 
     const name = camelCase(`${operation.getOperationId()}QueryOptions`)
 
-    const generics = [`TData = ${schemas.response.name}`, `TError = ${errors.map((error) => error.name).join(' | ') || 'unknown'}`]
+    const generics = [
+      `TData = ${schemas.response.name}`,
+      `TError = ${errors.map((error) => error.name).join(' | ') || 'unknown'}`,
+    ]
     const clientGenerics = ['TData', 'TError']
     const queryGenerics = [dataReturnType === 'data' ? 'TData' : 'ResponseConfig<TData>', 'TError']
+
+    console.log('schemas.queryParams?.name:' + schemas.queryParams?.name)
+    console.log('schemas.pathParams:')
+    console.log(schemas.pathParams)
     const params = createFunctionParams([
-      ...getDataParams(schemas.pathParams, { typed: true }),
+      ...getDataParams(schemas.pathParams, { typed: true })
+        .map(param => ({ ...param, required: false })),
       {
         name: 'params',
         type: schemas.queryParams?.name,
         enabled: !!schemas.queryParams?.name,
-        required: !!schemas.queryParams?.schema.required?.length,
+        required: false,
       },
       {
         name: 'headers',
         type: schemas.headerParams?.name,
         enabled: !!schemas.headerParams?.name,
-        required: !!schemas.headerParams?.schema.required?.length,
+        required: false,
       },
       {
         name: 'options',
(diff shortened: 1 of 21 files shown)


packages/swagger-swr/src/builders/QueryBuilder.ts
```diff
@@ -76,29 +84,34 @@ export function ${name} <${generics.join(', ')}>(${params}): SWRConfiguration<${
 
     const comments = getComments(operation)
 
-    const generics = [`TData = ${schemas.response.name}`, `TError = ${errors.map((error) => error.name).join(' | ') || 'unknown'}`]
+    const generics = [
```
stijnvanhulle, line 93: @b6pzeusbc54tvhw5jgpyw8pwz2x6gs Is there a reason why ``TError = ${errors.map((error) => error.name).join(' | ') || 'unknown'}`` has been removed?

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle Can you run the rest of the CI workflow that requires approval?

If it passes, I'll ask you to review my code as well.

stijnvanhulle:
@all-contributors please add @b6pzeusbc54tvhw5jgpyw8pwz2x6gs for SWR shouldFetch update

packages/swagger-swr/src/builders/QueryBuilder.ts
```diff
@@ -24,22 +24,26 @@ export class QueryBuilder extends OasBuilder<Config> {
...
     const params = createFunctionParams([
-      ...getDataParams(schemas.pathParams, { typed: true }),
+      ...getDataParams(schemas.pathParams, { typed: true })
+        .map(param => ({ ...param, required: false })),
       {
         name: 'params',
         type: schemas.queryParams?.name,
         enabled: !!schemas.queryParams?.name,
-        required: !!schemas.queryParams?.schema.required?.length,
+        required: false,
```
stijnvanhulle, line 44: @b6pzeusbc54tvhw5jgpyw8pwz2x6gs Any reason why `required` needs to be false. Normally this required is getting used to specify an order for the arguments.

```
const queryParams = createFunctionParams([
      ...getDataParams(schemas.pathParams, { typed: false }),
      {
        name: 'params',
        enabled: !!schemas.queryParams?.name,
        required: !!schemas.queryParams?.schema.required?.length,
      },
      {
        name: 'headers',
        enabled: !!schemas.headerParams?.name,
        required: !!schemas.headerParams?.schema.required?.length,
      },
      {
        name: 'clientOptions',
        required: false,
      },
])
```

And because we still have the previous check on line 100, it can cause some weird behaviours(e2e fails because of arguments changing places) 
Screenshot 2023-10-10 at 5 46 47 PM

stijnvanhulle (approved):

packages/swagger-swr/src/builders/QueryBuilder.ts
b6pzeusbc54tvhw5jgpyw8pwz2x6gs: @stijnvanhulle It's already been merged, but I'll answer it. I think it's better to allow React to not pass a value when it doesn't immediately know what value it needs for every parameter, rather than trying to pass meaningless fake values.

For example

```js
const token = useToken()
const userInfo = useGetUserInfo(token ? { token } : undefined, { shouldFetch: !!token })
```

When parameters are required, you should do it like below, but it is inconvenient to use `|| ''` for string type.

```js
const userInfo = useGetUserInfo({ token: token || '' }, { shouldFetch: !!token })
```

packages/swagger-swr/src/builders/QueryBuilder.ts
b6pzeusbc54tvhw5jgpyw8pwz2x6gs: I just added the newlines for simplicity of code style.

PR #483, opened 2023-10-13
feat(client,swr,tanstack-query): add 'clientImportPath' option

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
To support for referencing `client` with absolute or alias paths in the SWR plugin, add `clientImportPath` and deprectae `client` option.  The same applies to all three plugins: `@kubb/swagger-client`, `@kubb/swagger-swr`, and `@kubb/swagger-tanstack-query`.

The `client` option is not modified to maintain backward compatibility, and you can delete the code for major version updates later.

For example, you can make the following settings:

```js
createSwaggerSWR({
  output: './generated-api',
  clientImportPath: '@myalias-path/my-own-client',
}),
```


stijnvanhulle:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs 
Can you also adapt this to `@kubb/swagger-tanstack-query` and `@kubb/swagger-client`(same files but other directory)
```
clientPath ? getRelativePath(hook.filePath, clientPath) : '@kubb/swagger-client/client',
```

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle `swagger-client` has a behavior that copies the client file, while `swagger-swr` does not. For consistency, I would like both plugins to have the same behavior. Which do you prefer?

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle I am waiting for the answer to the above question.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle I am waiting for the answer to the above question.

I think it looks better not to copy the file.

stijnvanhulle:
> @stijnvanhulle `swagger-client` has a behavior that copies the client file, while `swagger-swr` does not. For consistency, I would like both plugins to have the same behavior. Which do you prefer?

It will probably be better to remove the copy client from the `swagger-client` package and use the `client` option as a reference/path. This will remove the confusion of which client is being used(if we copy one to the gen folder for example).

stijnvanhulle (approved):

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
When will this PR be merged?

PR #488, opened 2023-10-13
fix: correct parameters schema for `Zodios`(use of .shape and .schema.shape)

stijnvanhulle:
With a small change, it will generate the following(instead of using the full `Response` object):

**./zod/getUserSchema**
```
import { z } from 'zod'

export const getUserPathParamsSchema = z.object({ userId: z.string().describe(`Executes the action in the context of the specified user.`) })

/**
 * @description Successful response
 */
export const getUserQueryResponseSchema = z.object({ message: z.string().optional(), user: z.object({ userId: z.string().optional() }).optional() })

```

**zodios.ts**
```
import { makeApi, Zodios } from '@zodios/core'
import { getUserQueryResponseSchema, getUserPathParamsSchema } from './zod/getUserSchema'

const endpoints = makeApi([
  {
    method: 'get',
    path: '/users/:userId',
    description: `...`,
    requestFormat: 'json',
    parameters: [
      {
        name: 'userId',
        description: `Executes the action in the context of the specified user.`,
        type: 'Path',
        schema: getUserPathParamsSchema.shape['userId'],
      },
    ],
    response: getUserQueryResponseSchema,
    errors: [],
  },
])

export const api = new Zodios(endpoints)

export default api


```

### DIFF
--- a/examples/advanced/src/gen/zodios.ts
+++ b/examples/advanced/src/gen/zodios.ts
@@ -3,12 +3,25 @@ import {
   createPetsMutationResponseSchema,
   createPetsPathParamsSchema,
   createPetsQueryParamsSchema,
+  createPetsMutationRequestSchema,
+  createPetsHeaderParamsSchema,
   createPets201Schema,
 } from './zod/petsController/createPetsSchema'
-import { addPetMutationResponseSchema, addPet405Schema } from './zod/petController/addPetSchema'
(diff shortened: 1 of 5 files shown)


meister03 (approved):
Nice, you also added the description field. I will give it a test...

meister03:
Okay works on my test repo, thanks for the changes.

PR #493, opened 2023-10-16
feat(swagger-ts): support nullable types

raveclassic:
Hi, this PR adds support for `nullable: true` annotations for types in swagger-ts plugin. The annotation seems to be supported but there's a bug that skips creation of union type.


stijnvanhulle:
@all-contributors please add @raveclassic for code

raveclassic:
@stijnvanhulle I fixed the test but it seems there's a type issue somewhere in the internals, as `TypeBuilder#getTypeFromSchema` received `undefined` instead of object, I couldn't find the reason.

stijnvanhulle:
> Hi @stijnvanhulle, it seems something is off with E2E tests. `typecheck` shows an error which I can't reproduce locally. The issue is on the line `e2e/src/gen/mocks/databasesController/createDatabasesUpdateMaintenanceWindow.ts:56`, and TS in the current E2E CI snapshot is unable to assign return type of `createDatabaseMaintenanceWindow`, which is `DatabaseMaintenanceWindow` (or `{...} | null`) to `DatabasesUpdateMaintenanceWindowMutationRequest` which is `DatabaseMaintenanceWindow | null` (or `{...} | null | null`). Essentially, there should be no error.

Hi, can you pull in the main in to your branch? There was an issue on a previous PR but it should be fixed now.

PR #517, opened 2023-10-20
chore(swagger-client): copy client.ts only when the client options is provided

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
- With the `clientImportPath` option coming in, copying the client.ts file is only necessary when using the `client` option, so I modified the conditional statement.
- The existing clientPath was a string type, so the check was unnecessary.


b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle I just resolve conflicts, please code review this PR :)

stijnvanhulle (approved):

PR #553, opened 2023-10-25
Fix/path params type object

stijnvanhulle:


packages/swagger-client/src/generators/__snapshots__/OperationGenerator.test.ts.snap
```diff
@@ -0,0 +1,42 @@
...
+"
+`;
+
+exports[`OperationGenerator > [GET] should generate code based on a pathParamsType \`object\` 1`] = `
+"/**
+ * @summary Info for a specific pet
+ * @link /pets/:pet_id
+ */
+export async function showPetById<TData = ShowPetByIdQueryResponse>(
+  { petId, testId }: ShowPetByIdPathParams,
```
WickyNilliams, line 30: i kinda feel that the param names should be preserved verbatim from the schema in the object case. otherwise you have to understand not just that the names are converted to camelCase, but also how it handles the conversion for weird cases like e.g. `a_b_c`, `A_B_CD`, `_A__B_C` etc. the casing may have semantic meaning in the context of an API and this gets lost in the conversion.

WickyNilliams: do query params get converted to camelCase?

WickyNilliams:
thanks for the quick fix btw!

PR #559, opened 2023-10-26
Correctly construct LiteralTypes for properties that are "const"

helt:
Fixes #558

PR #583, opened 2023-11-01
More elaborate tests on const keyword

helt:
Although touched in #559 to be able to discriminate on types, the implementation is not in line with the spec for `const` itself.

Specifically, to adhere to the [json schema specification section about const](https://json-schema.org/draft/2020-12/json-schema-validation#name-const), const needs to take precedence over the `type` of a property. 

I also added a failing test, since I have no clue how to deal with it properly... 

@stijnvanhulle How shall we continue? (btw husky pre-commit hooks would keep the format clean...)


helt:
Closing because it becomes unmanagable

PR #662, opened 2023-11-22
Feature/more tests for corner cases

helt:
I moved const constrained detection further up the priorities, because to me it seems that const has a higher priority than most other property constraints. However, I have not found binding information about priorities in this regard - or if a property containing both, type and const, is an invalid openapi specification.

My assumption is that type constraints should be as rigid as possible.


stijnvanhulle:
hmm not sure how that is specified when following the spec. One thing that has been changed in our `beta` channel is doing an extra check if the OpenAPI version is v3 or v3.1. In the case of `const` this is only usable with v3.1.

stijnvanhulle (approved):

PR #702, opened 2023-12-04
fix: use of array with params when queryParams.name exists

stijnvanhulle:


examples/advanced/src/gen/clients/swr/petSWRController/useAddPet.ts
```diff
@@ -9,13 +9,13 @@ import type { AddPetMutationRequest, AddPetMutationResponse, AddPet405 } from '.
...
+  mutation?: SWRMutationConfiguration<ResponseConfig<TData>, TError>
   client?: Partial<Parameters<typeof client<TData, TError, TVariables>>[0]>
   shouldFetch?: boolean
-}): SWRMutationResponse<ResponseConfig<TData>, TError, string | null, TVariables> {
+}): SWRMutationResponse<ResponseConfig<TData>, TError> {
   const { mutation: mutationOptions, client: clientOptions = {}, shouldFetch = true } = options ?? {}
-  const url = shouldFetch ? `/pet` : null
-  return useSWRMutation<ResponseConfig<TData>, TError, string | null, TVariables>(url, (url, { arg: data }) => {
+  const url = `/pet` as const
+  return useSWRMutation<ResponseConfig<TData>, TError, typeof url | null>(shouldFetch ? url : null, (_url, { arg: data }) => {
```
0xemc, line 18: Should mutations also worry about params for keys?

stijnvanhulle:
```typescript
/**
     * The key of the resource that will be mutated. It should be the same key
     * used in the `useSWR` hook so SWR can handle revalidation and race
     * conditions for that resource.
     */

type ArgumentsTuple = readonly [any, ...unknown[]];
type Arguments = string | ArgumentsTuple | Record<any, any> | null | undefined | false;
type Key = Arguments | (() => Arguments);
```
Those are the types of key(first parameter) so yes it can use the params as key(same as being done in react-query).

PR #735, opened 2023-12-20
732 reduce amount of used generics

xPhentom:


packages/swagger-tanstack-query/src/components/__snapshots__/Mutation.test.tsx.snap
```diff
@@ -41,7 +41,7 @@ export function CreatePetsQueryOptions<TData = CreatePets['response'], TQueryDat
  * @summary Create a pet
  * @link /pets */
 
-export function useCreatePets<TData = CreatePets['response'], TQueryData = CreatePets['response'], TQueryKey extends QueryKey = CreatePetsQueryKey>(
```
stijnvanhulle, line 44: We still need data and queryKey, without you can not set `select` with correct types

PR #749, opened 2023-12-25
chore(swagger-zodios): export endpoints too

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
Sometimes it is necessary to use a dynamic baseUrl in a single build.

In this case, it is necessary to create a `Zodios` instance externally, as shown in the example code below, and therefore it is necessary to export the necessary endpoints.

```ts
import { Zodios } from "@zodios/core";
import { endpoints } from '@generated-api/zodios'

const API_SERVER_URL = process.env.API_SERVER_URL
const myAPI = new Zodios(API_SERVER_URL, endpoints)
```

I also added getAPI, which allows you to create an API instance without endpoints import.

```ts
import { getAPI } from '@generated-api/zodios'

const myAPI = getAPI(process.env.API_SERVER_URL)
```

### DIFF
--- a/packages/swagger-zodios/src/components/Definitions.tsx
+++ b/packages/swagger-zodios/src/components/Definitions.tsx
@@ -27,7 +27,8 @@ function Template({
 }: TemplateProps): ReactNode {
   return (
     <>
-      {`const endpoints = makeApi([${definitions.join(',')}])`}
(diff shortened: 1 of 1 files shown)


b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@stijnvanhulle Could you please release this PR?

PR #777, opened 2024-01-16
`format` attribute overrules `type` attribute in certain cases

helt:
This originated in the observation that  a plain file upload leads to a 

```yaml
    Plain_file:
      format: binary
      title: File
      type: string
```

```
export type PlainFile = string;
```
although a "Blob" would be more helpful in my case.


stijnvanhulle:
@helt What is the status of this pr? Seems like a nice to have bugfix for Kubb :)

helt:
@stijnvanhulle This is not done yet. I need feedback before this can be wrapped up and merged, I think.

I have to close this PR and opend #831. There are also open questions.

PR #781, opened 2024-01-18
fix: initialPageParam with type unknown to support string, number, ...

stijnvanhulle:
Infinite query page parameter is not always a number

### DIFF
--- a/.changeset/ten-windows-joke.md
+++ b/.changeset/ten-windows-joke.md
@@ -1,0 +1,5 @@
+---
+"@kubb/swagger-tanstack-query": patch
+---
+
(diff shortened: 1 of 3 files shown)


mpetito:
I'm not sure if there's an easy way to test this, but is the intention that the kubb configuration should still contain something like the following?

```
initialPageParam: 'undefined as string | undefined'
```

Otherwise, if I don't provide explicit typing, the type of `initialPageParam` is determined to be `undefined` when it should be `string | undefined`.  This creates a typing issue for the `queryFn` parameter `pageParam`.

image

PR #831, opened 2024-02-24
Give format precendence over type property

helt:
Equal to  #777 but rebased on current kubb main branch.

# Observation 

The JSON schema spec suggests that the `format` keyword has a higher specificity than the `type` keyword, since `format` it provides information about the contents of the schema at hand. 

Prior this PR, the `format` property would only be used by kubb iff the `type` property is not known. However, this prevents the `format` property to be used in most cases (except for `date`,  and `date-time`).

For example 
a plain file upload 

    Plain_file:
      format: binary
      title: File
      type: string

leads to a

    export type PlainFile = string;

although a `Blob` oder `FormData` would be more helpful in many cases.

# Goals

This PR enables `format` to "overrule" `type` and therefore generalizes the ad hoc approach that was begun `date` and `date-time` format parsing.

# Open questions:

However, some changes are still open for discussion, since they might involve inclusion of further dependencies (e.g. uuid validation).

* How to work with `format: uuid`?
* Should  the `format` only be used for `schema.type === "string"`, or shall it stay above those?
* is this a minor or a breaking change? Without activation flag, I would argue for breaking.
* Should this PR also trickle into downstream packages like zod or swr?

### DIFF
--- a/packages/swagger-ts/src/TypeGenerator.ts
+++ b/packages/swagger-ts/src/TypeGenerator.ts
@@ -1,17 +1,17 @@
-import { Generator } from '@kubb/core'
+import {Generator} from '@kubb/core'
 import transformers from '@kubb/core/transformers'
-import { getUniqueName } from '@kubb/core/utils'
+import {getUniqueName} from '@kubb/core/utils'
 import * as factory from '@kubb/parser/factory'
-import { keywordTypeNodes } from '@kubb/parser/factory'
-import { getSchemaFactory, isReference } from '@kubb/swagger/utils'
+import {keywordTypeNodes} from '@kubb/parser/factory'
+import {getSchemaFactory, isReference} from '@kubb/swagger/utils'
 
-import { pluginKey } from './plugin.ts'
+import {pluginKey} from './plugin.ts'
 
-import type { PluginManager } from '@kubb/core'
-import type { ts } from '@kubb/parser'
-import type { ImportMeta, Refs } from '@kubb/swagger'
-import type { Oas, OasTypes, OpenAPIV3, OpenAPIV3_1 } from '@kubb/swagger/oas'
-import type { PluginOptions } from './types.ts'
+import type {PluginManager} from '@kubb/core'
+import type {ts} from '@kubb/parser'
+import type {ImportMeta, Refs} from '@kubb/swagger'
+import type {Oas, OasTypes, OpenAPIV3, OpenAPIV3_1} from '@kubb/swagger/oas'
+import type {PluginOptions} from './types.ts'
 
 // based on https://github.com/cellular/oazapfts/blob/7ba226ebb15374e8483cc53e7532f1663179a22c/src/codegen/generate.ts#L398
 
@@ -19,6 +19,7 @@ type Context = {
   oas: Oas
   pluginManager: PluginManager
 }
+
 
 export class TypeGenerator extends Generator<PluginOptions['resolvedOptions'], Context> {
   refs: Refs = {}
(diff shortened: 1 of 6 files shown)


packages/swagger-ts/src/TypeGenerator.ts
```diff
@@ -386,8 +400,47 @@ export class TypeGenerator extends Generator<PluginOptions['resolvedOptions'], C
...
+        switch (schema.format) {
+          case 'binary':
+            if (schema.type === 'string') {
+              return factory.createTypeReferenceNode('Blob', [])
+            }
+            break
+          // 7.3.1. Dates, Times, and Duration
+          case 'date-time':
+          case 'date':
+          case 'time':
```
helt, line 422: actually, im not sure if `time` should really be cast a `Date`. Maybe `number` instead?

packages/swagger-ts/src/TypeGenerator.ts
```diff
@@ -386,8 +400,47 @@ export class TypeGenerator extends Generator<PluginOptions['resolvedOptions'], C
...
+            if (this.options.dateType === 'date') {
+              return factory.createTypeReferenceNode(factory.createIdentifier('Date'))
+            }
+          case 'uuid':
+            // TODO how to work with this? use https://www.npmjs.com/package/uuid for it?
```
helt, line 427: Currently, `uuid` would be ignored like all the others, although it would made sense to make it an `uuid`, but this would require a custom task or a dependency `uuid`

helt (commented):
@stijnvanhulle RFC

packages/swagger-ts/src/TypeGenerator.ts
```diff
@@ -386,8 +400,47 @@ export class TypeGenerator extends Generator<PluginOptions['resolvedOptions'], C
...
+          case 'binary':
+            if (schema.type === 'string') {
+              return factory.createTypeReferenceNode('Blob', [])
+            }
+            break
+          // 7.3.1. Dates, Times, and Duration
+          case 'date-time':
+          case 'date':
+          case 'time':
+            if (this.options.dateType === 'date') {
```
helt, line 423: How to write a test with this option activated?

stijnvanhulle:
How to work with format: uuid?
> I would say to ignore for the TS plugin(string would be fine) but for the Zod plugin we can use z.uuid

Should the format only be used for schema.type === "string", or shall it stay above those?

> Following the rules of JSON/OpenAPI it seems that format has a higher prior so should be checked first, same for other types

is this a minor or a breaking change? Without activation flag, I would argue for breaking.

> Jeah seems like that would change some types so I was thinking about using a new flag/option `strict` so we can change all of this based on strict mode.

 Should this PR also trickle into downstream packages like zod or swr?

> We will indeed also need those changes for Zod and Faker but I was also thinking about combining the logic between the 3 plugins so we only need to write the logic once but have not yet found the time to do so.

packages/swagger-ts/src/TypeGenerator.ts
stijnvanhulle: In the end, it will be a `string`, Typescript does not have a type uuid so I don't think we need to change something here for the ts plugin.

packages/swagger-ts/src/TypeGenerator.ts
stijnvanhulle: We have `dateType` which can now be a date or a string so maybe we can also add number if that would be needed for other users.

packages/swagger-ts/src/TypeGenerator.ts
helt: The thing is, the cases added here are directly from the openapi spec. That's why I added them here, too. 
But I'll remove unsupported cases if you like. 

Imo, being able to map to a specific type is very convinient. But kubb itself should add as little runtime dependencies as possible. Hence, uuid support via `@types/uuid` and `uuid` would be a great use case for a kubb plugin? Is that already possible with plugins, then I might give them a try.

vchirikov:
I'm also interesting in correct generation to `FormData` / `Blob` for:

```ts
      properties:
        file:
          type: string
          format: binary
```

@stijnvanhulle What I can use as workaround for now?

stijnvanhulle:
Moved to https://github.com/kubb-project/kubb/pull/935

PR #1017, opened 2024-05-28
Feature/fix zod tuple length

helt:
This is just a repro regarding the discovered tuples issue


helt:
Fixed by @stijnvanhulle himself

PR #1034, opened 2024-06-15
Mark enum as nullable for null enum members

### LINKED ISSUE #1030: Enum with null member crashes

2.19.5

Mac

_No response_

A enum with a member of `null` causes a crash. You can see it in action in the [playground](https://kubb.dev/playground) by pasting in the document below. To be clear, I want null as in the null type, not a string of the text "null".

Every time

My understanding is that [this is the correct way to model a nullable enum](https://swagger.io/docs/specification/data-models/enums/). I know that kubb will make a nullable enum if you set the type to null (eg `type: [string, "null"]`), but I am also using another OpenAPI tool that works a differently and right now these two tools are fighting each other. It seems like kubb is the right place to fix this since it's not adhering to the spec.

### Swagger/OpenAPI file?

```yaml
openapi: 3.1.0

components:
  schemas:
    Enum:
      enum:
        - a
        - b
        - c
        - null
```

### Additional information

Running locally, I get this error:

```
[8:34:31 PM]  ERROR  Cannot read properties of null (reading 'toString') (plugin: plugin-zod, hook: buildStart) (plugin: plugin-zod, hook: buildStart)

  at EventEmitter.<anonymous> (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/logger.ts:74:19)
  at EventEmitter.emit (node:events:518:28)
  at EventEmitter.emit (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/utils/EventEmitter.ts:10:19)
  at Object.emit (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/logger.ts:89:21)
  at PluginManager.catcher_fn (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/PluginManager.ts:620:17)
  at /home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/PluginManager.ts:366:9
  at Array.forEach (<anonymous>)
  at PluginManager.hookParallel (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/PluginManager.ts:362:13)
  at safeBuild (/home/apps/node_modules/.pnpm/@kubb+core@2.19.5/node_modules/@kubb/core/src/build.ts:208:5)
  at generate (/home/apps/node_modules/.pnpm/@kubb+cli@2.19.5_typescript@5.4.5/node_modules/@kubb/cli/src/generate.ts:79:36)
```

aburgel:
Fixes #1030 

According to the [3.0 docs](https://swagger.io/docs/specification/data-models/enums/) and [this FAQ](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md#if-a-schema-specifies-nullable-true-and-enum-1-2-3-does-that-schema-allow-null-values-see-1900), an enum is nullable when both it's type is nullable and when null is a member of the enum. Currently kubb will crash with a null member, so this PR fixes this case.

I added a couple of test cases as well, but let me know if there are more ways I should test this.

### DIFF
--- a/packages/plugin-oas/src/SchemaGenerator.ts
+++ b/packages/plugin-oas/src/SchemaGenerator.ts
@@ -560,6 +560,12 @@ export abstract class SchemaGenerator<
         type: 'type',
       })
 
+      const nullableEnum = schema.enum.includes(null);
+      if (nullableEnum) {
+        baseItems.push({ keyword: schemaKeywords.nullable })
+      }
+      const filteredValues = schema.enum.filter((value) => value !== null);
+
       // x-enumNames has priority
       const extensionEnums = ['x-enumNames', 'x-enum-varnames']
         .filter((extensionKey) => extensionKey in schema)
@@ -600,7 +606,7 @@ export abstract class SchemaGenerator<
                     value,
                     format: 'number',
                   }))
-                : [...new Set(schema.enum)].map((value: string) => {
+                : [...new Set(filteredValues)].map((value: string) => {
                     return {
                       name: value,
                       value,
@@ -624,7 +630,7 @@ export abstract class SchemaGenerator<
             name: enumName,
             typeName,
             asConst: false,
-            items: [...new Set(schema.enum)].map((value: string) => ({
+            items: [...new Set(filteredValues)].map((value: string) => ({
               name: transformers.stringify(value),
               value,
               format: isNumber(value) ? 'number' : 'string',
--- a/packages/swagger-ts/mocks/enums.yaml
+++ b/packages/swagger-ts/mocks/enums.yaml
@@ -1,91 +1,105 @@
-
-definitions:
-  enumVarNames.Type:
-    enum:
-    - 0
-    - 1
-    type: integer
-    x-enum-varnames:
-    - Pending
-    - Received
-  enumNames.Type:
-    enum:
-    - 0
-    - 1
-    x-enumNames:
+components:
+  schemas:
+    enumVarNames.Type:
+      enum:
+      - 0
+      - 1
+      type: integer
+      x-enum-varnames:
       - Pending
       - Received
-  enum.String:
-    type: string
-    enum:
-      - created_at
-      - description
-      - FILE.UPLOADED
-      - FILE.DOWNLOADED
-  enum.InObject:
-    type: object
-    properties:
-      reason:
-        type: string
-        enum:
-          - created_at
-          - description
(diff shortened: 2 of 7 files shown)


packages/plugin-oas/src/SchemaGenerator.ts
```diff
@@ -560,6 +560,12 @@ export abstract class SchemaGenerator<
         type: 'type',
       })
 
+      const nullableEnum = schema.enum.includes(null);
+      if (nullableEnum) {
+        baseItems.push({ keyword: schemaKeywords.nullable })
+      }
+      const filteredValues = schema.enum.filter((value) => value !== null);
```
aburgel, line 567: This is the fix. If `null` is one of the enum members, then the whole enum should be consider nullable. But we don't want `null` as one of the members, since that's not how Typescript works, so I filter it out.

packages/plugin-oas/src/SchemaGenerator.ts
```diff
@@ -560,6 +560,12 @@ export abstract class SchemaGenerator<
...
+
       // x-enumNames has priority
       const extensionEnums = ['x-enumNames', 'x-enum-varnames']
```
aburgel, line 570: I wasn't sure if I should use the filtered values in this case. Any thoughts on how to handle this?

packages/swagger-ts/mocks/enums.yaml
```diff
@@ -1,91 +1,105 @@
-
-definitions:
```
aburgel, line 2: I updated these test cases to OAS 3.1 since this new enum behavior doesn't appear in 2.0

packages/swagger-ts/mocks/enums_2.0.yaml
```diff
@@ -0,0 +1,23 @@
+definitions:
```
aburgel, line 1: I left this as swagger 2.0 because I wasn't sure if it was valid 3.1

stijnvanhulle:
@all-contributors add @aburgel for development

stijnvanhulle:
https://github.com/all-contributors add @aburgel for code

stijnvanhulle:
@all-contributors add @aburgel for code

PR #1054, opened 2024-06-25
Fix react-query dependency lookup

dgarciamuria:
# Issue
If you have multiple @tanstack packages in your package.json file, the logic that checks if your are using v5 will fail because internally is calling `isValidSync(/@tanstack/, '>=5')`

# Solution
Call `isValidSync` explicitly passing `@tanstack/react-query`

### DIFF
--- a/.changeset/kind-lions-check.md
+++ b/.changeset/kind-lions-check.md
@@ -1,0 +1,5 @@
+---
+"@kubb/swagger-tanstack-query": patch
+---
+
+Fix react-query dependency lookup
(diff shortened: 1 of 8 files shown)


stijnvanhulle:
Thanks but this will not work for vue-query, solid-query and such. Can you adapt the regex so that `react` is not hardcoded?

stijnvanhulle:
@all-contributors add @dgarciamuria for code

dgarciamuria:
> Thanks but this will not work for vue-query, solid-query and such. Can you adapt the regex so that `react` is not hardcoded?

Good catch! I decided to be explicit instead of trying to form of fuzzy matching, the downside is that if react-query adds support for a new framework then the regex will need to be updated...but is better than risking capturing other packages that are not related.
Also moved the regex as an export to make it easier to refactor in the future

PR #1055, opened 2024-06-25
Fix/object type

dgarciamuria:
# Issue
If you have `typescript-eslint` enabled in your project, you might run into a situation where you get some errors related about the use of `{}` as type after generating the code.

# Solution
This adds a check to see if `nodes.length` is 0, meaning the node doesnt have any children and thus should be treated as an object instead of creating a new literal node...probably the original condition was wrong since `![]` is a falsy value, but decided to keep it as it was to avoid any potential regression

### DIFF
--- a/.changeset/big-hotels-join.md
+++ b/.changeset/big-hotels-join.md
@@ -1,0 +1,5 @@
+---
+"@kubb/swagger-ts": patch
+---
+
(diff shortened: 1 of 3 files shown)


stijnvanhulle:
@all-contributors add @dgarciamuria for code

dgarciamuria:
updated snapshot

PR #1148, opened 2024-08-27
isTypeOnly set to true for all generated schema's(typescript)

stijnvanhulle:


tbjgolden:
I think the update to execa's version added a dependency on

`import {addAbortListener} from 'node:events';` which was added in 18.18, just fyi for the supported node engine versions

PR #1166, opened 2024-08-29
Integer validation schema not generated properly

stijnvanhulle:
I noticed in [SchemaGenerator](https://github.com/kubb-project/kubb/blob/e97679edbd576d5b38a996d69d7835d51fd8f874/packages/plugin-oas/src/SchemaGenerator.ts) that when a property has default set we only did a check on string(typeof string) and boolean(typeof boolean). All other formats(like in this case number) were ignored.


packages/swagger-zod/src/components/__snapshots__/Schema.test.tsx.snap
```diff
@@ -12,9 +12,21 @@ export const Order = z.object({"status": z.enum(["placed", "approved"])})
 "
 `;
 
+exports[`<Schema/>  > 'PageSizeInteger' 1`] = `
```
stijnvanhulle, line 15: @assassinoNz the test case for number and default

packages/plugin-oas/src/SchemaGenerator.ts
```diff
@@ -407,12 +401,16 @@ export abstract class SchemaGenerator<
...
           args: transformers.stringify(schema.default),
         })
-      }
-      if (typeof schema.default === 'boolean') {
+      } else if (typeof schema.default === 'boolean') {
         baseItems.push({
           keyword: schemaKeywords.default,
           args: schema.default ?? false,
         })
+      } else {
```
stijnvanhulle, line 409: @assassinoNz the missing part for default

PR #1176, opened 2024-08-31
Fixes/swr mutations key fn

helt:
The example oas provided in `e2e/schemas/dataset_api.yaml`  demonstrates that generation of swr mutations result in typescript errors. I observed this behaviour when the key (the rest path) contains parameters. 

These are fixed when the swr type `Key` is used instead. 

Interestingly, `pnpm generate && pnpm typecheck` in e2e does not raise this. So this might be a bug on its own?


helt:
fixes #1177

stijnvanhulle:
Thanks for fixing this. Weird that the typescript check is not catching this. Do you have a playground/example where this was failing, I tried to reproduce this but the swagger you provided does not crash in the latest Kubb config. Maybe a mismatch of versions(we are using the latest typescript and swr version).

stijnvanhulle (approved):

PR #1188, opened 2024-09-03
Fix strict e2e for swr

helt:


packages/plugin-swr/src/components/Mutation.tsx
```diff
@@ -137,17 +140,18 @@ export function Mutation({ factory, Template = defaultTemplates.default }: Props
...
-  const client = {
+  const client : TemplateProps["client"] = {
     method: operation.method,
     path: new URLPath(operation.path),
     generics: [`${factory.name}["data"]`, `${factory.name}["error"]`, schemas.request?.name ? `${factory.name}["request"]` : ''].filter(Boolean).join(', '),
     withQueryParams: !!schemas.queryParams?.name,
     withData: !!schemas.request?.name,
     withPathParams: !!schemas.pathParams?.name,
     withHeaders: !!schemas.headerParams?.name,
+    fetcherParameterization: `_url: Key${!!schemas.request?.name ? `, { arg: data } : {arg: ${factory.name}["request"]  }` : ''}`
```
helt: @stijnvanhulle I deem this awkward but it does the job (generate the parameters that belong inside the `()` brackets of the async fetcher function). Maybe you have an alternative approach thats nicer?

helt:
zod now fails due to:
```
Error: src/gen/zod/notificationFilterEmailSchema.ts(3,15): error TS7022: 'notificationFilterEmailSchema' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.
```` 
image

This stems from `bunq.com.json` schema, from the component below. Obviously, this looks like a mistake (by bunq), because this array element cant contain anything but itself... But it is what it is, syntactically correct oas...
```
      "NotificationFilterEmail": {
        "properties": {
          "notification_filters": {
            "description": "The types of notifications that will result in a email notification for this user.",
            "items": {
              "$ref": "#/components/schemas/NotificationFilterEmail"
            },
            "readOnly": false,
            "type": "array",
            "writeOnly": false
          }
        },
        "type": "object"
      },
```

File: packages/plugin-swr/src/components/Mutation.tsx (hunk 1 above)
Comment (stijnvanhulle) [reply packages/plugin-swr/src/components/Mutation.tsx]: That's fine, I have been working on a helper to create functionParameters but not fully finished yet. For example:

```
import { getFunctionParams } from '@kubb/react'

getFunctionParams({
  name: {
    type: 'string',
    optional: true,
  },
  age: {
    type: 'number',
    optional: true,
  },
  address: {
    type: 'string',
  },
}, { type: 'constructor' })
```
will result in `address: string, name?: string, age?: number`

**Comment by stijnvanhulle** [comment]
> zod now fails due to:
> 
> ```
> Error: src/gen/zod/notificationFilterEmailSchema.ts(3,15): error TS7022: 'notificationFilterEmailSchema' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.
> ```
> 
> image
> 
> This stems from `bunq.com.json` schema, from the component below. Obviously, this looks like a mistake (by bunq), because this array element cant contain anything but itself... But it is what it is, syntactically correct oas...
> 
> ```
>       "NotificationFilterEmail": {
>         "properties": {
>           "notification_filters": {
>             "description": "The types of notifications that will result in a email notification for this user.",
>             "items": {
>               "$ref": "#/components/schemas/NotificationFilterEmail"
>             },
>             "readOnly": false,
>             "type": "array",
>             "writeOnly": false
>           }
>         },
>         "type": "object"
>       },
> ```
> 
> https://zod.dev/?id=recursive-types

Hmm maybe we need to update that example, that has been a swagger/openAPI example of last year(they probably already resolved it). If not we can ignore it for now

File: e2e/kubb.config.js [hunk 2]
```diff
@@ -96,7 +96,7 @@ const baseConfig = {
         exportType: false,
       },
       group: { type: 'tag' },
-      typed: false,
+      typed: true,
```
helt: @stijnvanhulle this fixes the zod issue

PR #1208, opened 2024-09-09
Include 'null' in list of valid types

aburgel:
`type: "null"` is a valid type in OAS 3.1, but this line will result in a warning if it's used.


stijnvanhulle:
Ooh, I was not aware of that, thanks for fixing this. I will also merge this in the v3 branch(main) so v3 beta could also have this fix.

stijnvanhulle (approved):

PR #1250, opened 2024-09-23
fix: add missing freeform object definition to plugin-oas schema generator

### LINKED ISSUE #1232: freeform object is not created correctly in plugin-ts

2.17.0, but works the same with 3.0.0-alpha.20

macos

irrelevant

There are two ways to express a "freeform object" in OpenAPI: either add `additionalProperties: true` or `additionalProperties: {}` to the entity with type `object` ([source](https://swagger.io/docs/specification/data-models/dictionaries/#free-form)).

The `plugin-ts` does perfectly fine with former way and generates 
```typescript
freeformObject: {
    [key: string]: any
}
```
which is an accurate way to express an object with keys as strings (as per spec) and any kind of values.

The latter way of expressing the same, `additionalProperties: {}`, has a bug - it generates the following:
```typescript
freeformObject: object
```
which basically says that it can be anything except `null` or `undefined`, but the compiler doesn't know anything about the value, so doing something like `freeformObject['someRandomKey'] = 2` throws `Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'`.

None

get 
```typescript
freeformObject: {
    [key: string]: any
}
```
in both cases

### Swagger/OpenAPI file?

providing only the relevant part of the slightly changed petStore.yml you use in your project:
```yaml
Pet:
      required:
        - name
        - photoUrls
      type: object
      properties:
        id:
          type: integer
          format: int64
          example: 10
        name:
          type: string
          example: doggie
        freeformObject:
          type: object
          additionalProperties: {}
        category:
          $ref: 'https://petstore3.swagger.io/api/v3/openapi.json#/components/schemas/Category'
        photoUrls:
          type: array
          xml:
            wrapped: true
          items:
            type: string
            xml:
              name: photoUrl
        tags:
          type: array
          xml:
            wrapped: true
          items:
            $ref: '#/components/schemas/Tag'
        status:
          type: string
          description: pet status in the store
          enum:
            - available
            - pending
            - sold
      xml:
        name: pet
```

### Additional information

So I went on to dig into what was happening and ended up in `SchemaGenerator.ts` of `plugin-oas`. 
Found the place where it checks for `additionalProperties === true` and added `|| !Object.keys(additionalProperties).length`, hoping for an easy fix; however, it didn't change a thing. 

So I set a breakpoint there and found out that if `additionalProperties` contains `{}`, it gets stripped (the object that gets there is just `{type: 'object'}`, without `additionalProperties` at all); I went on as high as I saw inside the flow and got convinced to a certain degree that it gets stripped in `oas` during the initial parsing.

Unfortunately, I didn't have too much time to play with it, and the code there is not the simplest; if you could direct me to where else I can look to find out the root cause I'd be thankful. And thanks for the great project of course!! :)

grreeenn:
closes #1232

### DIFF
--- a/.changeset/poor-candles-wash.md
+++ b/.changeset/poor-candles-wash.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-oas": patch
+---
+
+Take care of turning `additionalProperties: {}` to a correct freeform object type (#1232)
--- a/packages/plugin-oas/src/SchemaGenerator.ts
+++ b/packages/plugin-oas/src/SchemaGenerator.ts
@@ -302,7 +302,7 @@ export class SchemaGenerator<
 
     if (additionalProperties) {
       additionalPropertiesSchemas =
-        additionalProperties === true
+        additionalProperties === true || !Object.keys(additionalProperties).length
           ? [{ keyword: this.#getUnknownReturn({ schema, name }) }]
           : this.parse({ schema: additionalProperties as SchemaObject, parentName: name })
     }


stijnvanhulle:
@grreeenn Thanks for having a look and opening a PR, really appreciate that. After the merge, I will cherry-pick so v2 could also have this fix.

stijnvanhulle:
@all-contributors please add @grreeenn for code

stijnvanhulle (approved):

PR #1282, opened 2024-10-03
feat: added option to coerce specific types in plugin-zod

### LINKED ISSUE #1281: Allow specifying which types to coerce [plugin-zod]

I'm using Kubb for a serverless API, and i use to to generate Zod-schemas from my OpenApiSpec. 

In this case I'm making use of the `coerce` feature of Zod. However this works better for some data-types than others.

In my case i would like to coerce for numbers (`z.coerce.number()`) and not strings, since they behave pretty badly [ref](https://github.com/colinhacks/zod/issues/2520).

Here's my current config

```typescript
import { defineConfig } from '@kubb/core';
import { pluginZod } from '@kubb/swagger-zod';

export default defineConfig({
  input: {
    path: './test/openapi-spec.yaml',
  },
  output: {
    path: './test/generated',
  },
  plugins: [
    pluginOas({
      output: {
        path: '.json-schemas',
      },
      validate: true,
    }),
    pluginZod({
      output: {
        path: 'schemas',
      },
      typedSchema: true,
      dateType: 'stringOffset',
      unknownType: 'unknown',
      coercion: true,
      importPath: 'zod',
    }),
  ],
});
```

I would like to add another option to the Zod plugin 
```typescript
export type Options = {
   ...
   coerceTypes: { // Default value would be true for all, to be backwards compatible
     date: boolean
     strings: boolean
     number: boolean
   },
   ...
}
```

And then use that in conjuction with the `coercion` option, to enabled coercion.

I'd be down to write the PR as well, if you accept this enhancement.

ChilloManiac:
Closes #1281 

I hope I've filled all the parts needed :). I was unsure if i should declare it as a major or minor or patch bump, as it seems you've kept things at 3.0.0 for the big release.


stijnvanhulle:
@all-contributors please add @ChilloManiac for code

docs/plugins/plugin-zod/index.md
```diff
@@ -201,6 +201,11 @@ z.string()
 z.date()
 z.number()
 ```
+```typescript [{numbers: true}]
+z.string()
```
stijnvanhulle, line 205: z.string()
z.date()
is not needed here + can you also add example for date and string?

packages/plugin-zod/src/types.ts
```diff
@@ -93,7 +100,7 @@ type ResolvedOptions = {
   mapper: NonNullable<Options['mapper']>
   importPath: NonNullable<Options['importPath']>
   coercion: NonNullable<Options['coercion']>
-  operations: NonNullable<Options['coercion']>
+  operations: NonNullable<Options['operations']>
```
stijnvanhulle, line 103: Thanks for fixing this, forgot about this one.

packages/plugin-zod/src/parser/index.ts
```diff
@@ -150,14 +150,25 @@ export function sort(items?: Schema[]): Schema[] {
   return transformers.orderBy(items, [(v) => order.indexOf(v.keyword)], ['asc'])
 }
 
+const shouldCoerce = (coercion: ParserOptions['coercion'] | undefined, type: 'dates' | 'strings' | 'numbers'): boolean => {
```
stijnvanhulle, line 153: Can you also a small test in https://github.com/kubb-project/kubb/blob/c8cb50d1e4a13669a05ca11a18352b86a558bce1/packages/plugin-zod/src/generators/zodGenerator.test.tsx

Appending something like this to the const testData will add another test case with those new options.
```
  {
      name: 'Pet',
      path: 'Pet',
      input: '../../mocks/petStore.yaml',
      options: {
           coercion: { ...}
     },
    },
```

stijnvanhulle (approved):

docs/plugins/plugin-zod/index.md
ChilloManiac: My intent was to show that it's configurable individually.

Which would you prefer?

A block like the above for each data type, or a single block like:
```
```typescript [{numbers: true, strings: false, dates: false}]
z.string()
z.date()
z.coerce.number()
```


where dates and strings are explicitly false?

ChilloManiac:
> @ChilloManiac Minor is fine, if needed we can cherry pick this to v2.

Cherry picking into v2 would be awesome, thanks 🚀

docs/plugins/plugin-zod/index.md
stijnvanhulle: Ooh that is also fine and less code in the docs. Your example here seems fine, thanks.

ChilloManiac: Added the example :)

PR #1292, opened 2024-10-04
fix: discard optional if default is present

### LINKED ISSUE #1290: Plugin-zod should discard "optional" if "default" is present

v2.26.3

MacOs

"@kubb/cli": "2.26.3",     "@kubb/core": "2.26.3",     "@kubb/plugin-oas": "2.26.3",     "@kubb/swagger": "2.26.3",     "@kubb/swagger-zod": "2.26.3",

Yaml file:
```yaml
openapi: 3.0.0

components:
  schemas:
     List:
        type: object
        properties: 
          limit:
            description: Upper limit of amount of items to be fetched
            type: integer
            minimum: 1
            default: 100
            example: 10
```
generated zod schema
```typescript
  z.object({
    limit: z.coerce
      .number()
      .int()
      .default(100)
      .describe('Limit of items returned in response.')
      .optional(),
  })
```

`optional` and `default` are not really used together, as  optional will make the output type (`z.infer`) be optional, which it never actually should be because of the default. Whereas `default` will only make the input type (`z.input`) be optional and not the output

Every time

Expected zod schema would be

```typescript
  z.object({
    limit: z.coerce
      .number()
      .int()
      .default(100)
      .describe('Limit of items returned in response.')
  })
```

### Swagger/OpenAPI file?

_No response_

### Additional information

Should be doable in [Zod.tsx](https://github.com/kubb-labs/kubb/blob/main/packages/plugin-zod/src/components/Zod.tsx#L20C1-L30C18)

by introducing a `const hasDefault = ....` and then adding to the filter `if (hasDefault && isKeyword(item, schemaKeywords.optional) return false`

ChilloManiac:
Closes #1290

### DIFF
--- a/.changeset/selfish-teachers-rhyme.md
+++ b/.changeset/selfish-teachers-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-zod": patch
+---
+
+Discard `optional()` if there is a `default()` to ensure the output type is not `T | undefined`
--- a/packages/plugin-zod/src/components/Zod.tsx
+++ b/packages/plugin-zod/src/components/Zod.tsx
@@ -19,11 +19,15 @@ type Props = {
 
 export function Zod({ name, typeName, tree, inferTypeName, mapper, coercion, keysToOmit, description }: Props): KubbNode {
   const hasTuple = tree.some((item) => isKeyword(item, schemaKeywords.tuple))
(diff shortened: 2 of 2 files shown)


ChilloManiac:
Looked through the typechecking, but can't figure out why errors would show up in vue and react query here 🤷‍♂️

stijnvanhulle (approved):

PR #1341, opened 2024-10-08
fix: discard optional if default inside nested objects

ChilloManiac:
This is a followup to #1292 as it did not work in nested objects

This PR adds the capability to discard `optional` in nested objects.

I'm not very satisfied with the solution, but i found it hard to find a way to manipulate the AST as it resolves to a string immediately. The issue is that we'd need to make a decision on wether to include the keyword `optional` depending on whether or not `default` shows up before or after. But from the perspective of the `current` keyword there's no way to know about it's siblings.

I didn't want to introduce a refactor, but im open to it if you have some agreement to it.

Two other ways to go about this
1. Introduce a `siblings: Schema[]` to `parse`.
2. Refactor `parse` to output some kind of tree-structure that can then be manipulated after the fact, and before reducing it to a string. Then the tree could be pruned of `optionals` using a Visitor

### DIFF
--- a/packages/plugin-zod/mocks/petStore.yaml
+++ b/packages/plugin-zod/mocks/petStore.yaml
@@ -20,6 +20,12 @@ paths:
           required: false
           schema:
             type: string
+        - name: offset
+          in: query
+          required: false
+          schema:
+            type: integer
+            default: 0
       responses:
         '200':
           description: A paged array of pets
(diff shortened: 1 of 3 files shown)


stijnvanhulle (approved):

ChilloManiac:
@stijnvanhulle I have a hunch i could get Option 2 working pretty quickly - Atleast for this plugin, haven't looked at the others at all 😄  I can throw together a PR pretty if you'd like (Before this is merged)

PR #1342, opened 2024-10-10
feat: siblings in parser-zod for better AST manipulation

ChilloManiac:
This is an alternative to #1341 and #1292 

I like this implementation better, but it might differ from the "usual" way plugins are structured in @kubb atm


packages/plugin-zod/src/parser/index.ts
```diff
@@ -171,7 +171,7 @@ type ParserOptions = {
   coercion?: boolean | { dates?: boolean; strings?: boolean; numbers?: boolean }
 }
 
-export function parse(parent: Schema | undefined, current: Schema, options: ParserOptions): string | undefined {
+export function parse(parent: Schema | undefined, current: Schema, options: ParserOptions, siblings: Schema[]): string | undefined {
```
stijnvanhulle: we can probably refactor this as follow:
(not sure about the name `ParserAst`)

```
type ParserAst=  {
   parent: Schema | undefined,
   current: Schema
   siblings: Schema[]
}
export function parse({ current, parent, siblings } : ParserAst , options: ParserOptions): string | undefined {
```

ChilloManiac: I've given it a go, in the newest version here

ChilloManiac:
@stijnvanhulle I've made the refactor in this 😊  Feel free to either merge this or #1341 depending on your preference. The other should be closed afterwards 👌

stijnvanhulle:
@ChilloManiac This looks better, I would prefer this kind of structure for the parser. 

We only need to move this to the other parsers(zod, ts and faker plugin) so we have the same way of getting siblings and the parent(like you have done here).

stijnvanhulle (approved):

ChilloManiac:
@stijnvanhulle will you do those changes? And will it be a part of this PR?

stijnvanhulle:
@ChilloManiac You can already merge this one and then when I find some time this or next week I can also do the same for the other plugins that uses parsers.

ChilloManiac:
> @ChilloManiac You can already merge this one and then when I find some time this or next week I can also do the same for the other plugins that uses parsers.

Sorry, but only those with write access to your repo can merge it, even if its approved 😅

PR #1360, opened 2024-10-29
Make 'Mutation' configs optional

CHE1RON:
This is the same as for `Query` etc, since otherwise users always have to specify `key` (which is not the case when specifying `Query` options, and users should not have to)


stijnvanhulle:
@all-contributors please add  @CHE1RON for code

CHE1RON:
Thanks very much! 😊

PR #1367, opened 2024-10-31
Support Arrays in multipart/form-data

shahriar-shojib:


gusthavosouza:
@shahriar-shojib can we make sure this will also work for file uploading?

I had to remove the code from my client generation as it was breaking file uploading:

`if(data) {
    Object.keys(data).forEach((key) => {
      const value = data[key as keyof typeof data];
      if (typeof key === "string" && (typeof value === "string" || value instanceof Blob)) {
        formData.append(key, value);
      }
    })
   }`

shahriar-shojib:
> @shahriar-shojib can we make sure this will also work for file uploading?
> 
> I had to remove the code from my client generation as it was breaking file uploading:
> 
> `if(data) { Object.keys(data).forEach((key) => { const value = data[key as keyof typeof data]; if (typeof key === "string" && (typeof value === "string" || value instanceof Blob)) { formData.append(key, value); } }) }`

I don't know why it would break because the code you sent is the existing code and it works for file uploads as long as the formData does not contain any Arrays.

I will double-check the test cases.

please add more details on how your file uploads are breaking so that I can take a closer look.

gusthavosouza:
@shahriar-shojib I think this will do the job.

My project we are using Expo with multiple file selection, and by default it was converting the form-data to array with the file reference, something like:

image

I do feel that this will solve the issue.

Thanks for your response

shahriar-shojib:
> @shahriar-shojib I think this will do the job.
> 
> My project we are using Expo with multiple file selection, and by default it was converting the form-data to array with the file reference, something like:
> 
> image
> 
> I do feel that this will solve the issue.
> 
> Thanks for your response

this is indeed what this PR aims to solve!

vitorcamachoo:
Any news on this? We have a critical bug in our company due to this issue.

shahriar-shojib:
@stijnvanhulle 👀

Thijmen:
@stijnvanhulle can this get some attention?

stijnvanhulle:
Opening another pr that will move this to one file for all use cases

Issue #1399, opened 2024-11-15
Zod generation is considering blob as a string


ekaradon:
### What version of `kubb` is running?

3.0.5

### What platform is your computer?

Linux

### What steps can reproduce the bug?

Just generate zod from an openAPI containing a file type. See the openAPI example below.

### How often does this bug happen?

Every time

### What is the expected behavior?

  blob: () => 'z.instanceof(File)',

### Swagger/OpenAPI file?

```
openapi: 3.0.3
info:
  contact:
    email: support@example.com
    name: API Support
    url: http://www.example.com/support
  description: API Spec for exports
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
  termsOfService: http://example.com/terms/
  title: View Exports
  version: 1.0.0
servers:
  - url: /exports/
paths:
  /pdf-export:
    post:
      description: Run a pdf export
      operationId: runPdfExport
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/RunPdfExportRequest"
      responses:
        "200":
          content:
            application/octet-stream:
              schema:
                format: binary
                type: file
          description: Run a pdf export response
        default:
          description: An error occurred while processing the request.
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ErrorResponse"
      tags:
        - Export
      x-content-type: application/json
      x-accepts: application/octet-stream
      x-tags:
        - tag: Export
components:
  parameters:
    container:
      explode: false
      in: header
      name: x-container-uuid
      required: true
      schema:
        type: string
      style: simple
    user:
      explode: false
      in: header
      name: x-user-uuid
      required: true
      schema:
        type: string
      style: simple
  schemas:
    ErrorResponse:
      type: object
      required:
        - status
        - message
      properties:
        status:
          type: integer
          format: int32
        message:
          type: string
    RunPdfExportRequest:
      properties:
        url:
          type: string
        timezone:
          type: string
        language:
          type: string
        site_id:
          type: string
        max_wait_millis:
          type: integer
        format:
          enum:
            - PDF
            - PNG
          type: string
      required:
        - format
        - language
        - site_id
        - timezone
        - url
      type: object
```

### Additional information

I can do the pull request if you want for just modifying the `index.ts` file in `packages/plugin-zod/src/parser/index.ts`.

stijnvanhulle:
[...]

PR #1400, opened 2024-11-15
Fix check for blob should check for a file instead of a string

ekaradon:
I've continued working with zod and got an error when using it to parse a response from the backend service where it tries to match a blob with a string. I suppose that `z.instanceOf(File)` would be the best approach here?

[KUBB-70 Zod generation is considering blob as a string](https://linear.app/kubb/issue/KUBB-70/zod-generation-is-considering-blob-as-a-string)
https://github.com/kubb-labs/kubb/issues/1399


stijnvanhulle:
@all-contributors please add @ekaradon for code

stijnvanhulle:
@ekaradon hmm I was not aware off that could use `z.instance` so seems more than logical to use `z.instanceOf(File)` here. Not sure if that would have an impact on people using strings for Blobs.

ekaradon:
I'm not used to workspace, so I may have missed a step but when I try to run tests after installing with `pnpm` I get this error message:

```
Error: Failed to resolve entry for package "@kubb/fs". The package may have incorrect main/module/exports specified in its package.json.
  Plugin: vite:import-analysis
  File: ~/projects/kubb/packages/core/src/utils/parser.ts
  ```
  
  Not sure what I should do to get it working?

PR #1478, opened 2024-12-18
chore: update `plugin-oas` examples to use empty generators

bohdanhusak:
Replace `output: false` with `generators: []` in `plugin-oas` options for all examples that use it.


stijnvanhulle:
@all-contributors please add @bohdanhusak for docs

PR #1563, opened 2025-02-15
feat(plugin-zod): wrapOutput option

msutkowski:
This is just an idea that would be helpful for one of my use cases. The TLDR is I have a gateway that takes N schemas, normalizes and/or transforms them, then recomposites them into normalized routes. One of the biggest pitfalls with every generator is that they don't account for examples in existing OpenAPI specs, so I end up having to generate things then extend them as well.

This would allow to keep the examples and not rely on whatever is serving the documentation to handle it. Especially useful for multiple examples.

~**A better API** for this might be to just support `wrapOutput` or `transformOutput`. There are libs/wrappers that work like this test I've added, as well as ones that behave like `extendApi(z.object({}), { example: "test", description: "example desc.", ...otherOpenApiProperties })`~

Thanks and very much open to any feedback :)

Note: if you're good with the idea, I'll update everything so it's fully ✅


stijnvanhulle:
@msutkowski Thanks for opening the pr, this was indeed not yet added to Kubb. The idea itself seems fine to me, you could finish this and update the `changelog.md` to include this change(https://github.com/kubb-labs/kubb/blob/main/docs/changelog.md).

stijnvanhulle (approved):

stijnvanhulle:
Could you also update https://github.com/kubb-labs/kubb/blob/main/docs/plugins/plugin-zod/index.md to include the new option `wrapOutput`?

msutkowski:
> Could you also update https://github.com/kubb-labs/kubb/blob/main/docs/plugins/plugin-zod/index.md to include the new option `wrapOutput`?

Updated. My 🧠 was having trouble with writing copy for this, but I might just write a small blog post or something to elaborate on it 😂. The use case on this is pretty niche, but it is valuable. Kubb is _great_ for easily building a microservices gateway where you might need to make slight modifications to schemas for normalization purposes, as well as generating clients for the underlying services. It'd make more sense if I pointed to specific libraries, but I was trying to avoid that and just stick to the concept.

CleanShot 2025-02-21 at 21 02 33@2x

stijnvanhulle:
@all-contributors please add @msutkowski  for code

PR #1572, opened 2025-02-22
fix: discriminator mapping with literal types

stijnvanhulle:


packages/plugin-ts/mocks/discriminator.yaml
```diff
@@ -205,5 +205,39 @@ components:
...
+    FooString:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/FooBase'
+      properties:
+        value:
+          type: string
+      required:
+        - value
+      additionalProperties: false
```
telmofcosta, line 230: Is this format allowed? Shouldn't it be
```yaml
    FooString:
      allOf:
        - $ref: '#/components/schemas/FooBase'
        - type: object
          properties:
            value:
              type: string
          required:
            - value
      additionalProperties: false
```

PR #1577, opened 2025-02-25
Support void as unknownType

vitorcamachoo:
We work in an API, and in some cases, it returns an empty payload (404, 202, etc.), and for those cases, we want the void type. Currently, the Kubb configuration only allows `unknown` and `any`.
We are migrating from Orval to Kubb, and for those situations, Orval handles them as `void` which is the most correct in our use case.

Example:
When using a union in Typescript, in which one of the types is unknown or any, the return type is unknown or any.


vitorcamachoo:
Please let me know if there is anything missing in the code.

stijnvanhulle:
@vitorcamachoo Thanks for opening a pr, really appreciate your work here.
Could you also update the types for `unknownType` in https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-zod/src/types.ts#L50-L50, https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-faker/src/types.ts#L53-L53 and https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-ts/src/types.ts#L58-L58.

And next to that, updating the docs to also include the new  `void` option.

stijnvanhulle:
@all-contributors please add @vitorcamachoo  for code

vitorcamachoo:
> @vitorcamachoo Thanks for opening a pr, really appreciate your work here. Could you also update the types for `unknownType` in https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-zod/src/types.ts#L50-L50, https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-faker/src/types.ts#L53-L53 and https://github.com/kubb-project/kubb/blob/1f199b882306088fa10b4b59662698ac45aeafca/packages/plugin-ts/src/types.ts#L58-L58.
> 
> And next to that, updating the docs to also include the new `void` option.

Thank you, @stijnvanhulle, for your feedback. I forgot to add the `void` in `plugin-faker/src/types.ts`, but it was already included in `plugin-ts/src/types.ts`.

stijnvanhulle (approved):

PR #1580, opened 2025-02-26
Cypress Plugin

### LINKED ISSUE #1579: Cypress plugin

Making requests with Cypress is not type-checked (neither the request nor the body), making it difficult to work with.

### External documents/projects?

- [Cypress request](https://docs.cypress.io/api/commands/request#__docusaurus_skipToContent_fallback)
- [Cypress custom commands](https://docs.cypress.io/api/cypress-api/custom-commands)

### What is the feature you are proposing to solve the problem?

It would be great if a plugin for Cypress was created that would allow generation of commands that accept only the right request body, and return a `Chainable<Response>`.

Something like : 
```ts
// CypressPetCreateRequest.ts

namespace Cypress {
   interface Chainable {
      createPetRequest: (body: CreatePetRequestBody) => Chainable<CreatePetResponse>
   }
}
```

### What alternatives have you considered?

For now I'm using `cy.request` directly.

Ericlm:
Implements the automatic generation of typed Cypress requests as Cypress commands.

Closes #1579


Ericlm:
I'm not sure to add the functions to the Cypress global commands as it could add a lot of `noise`. However, they will be `chainable` so users would be able to wrap the requests in commands if necessary.

Ericlm:
### Things to know
- This is my first contribution to this project (and in using `tsx`) so I guess I didn't follow the best practices
- Pretty much all of the plugin in based on plugin-msw
- While updating the documentation I thought it would be nice to have those in a `cypress` folder ; however I don't think there is anything in the plugin that says so

- I removed `fakerName` as it wasn't useful in this plugin
- I didn't make cypress commands as it could be annoying (but could be added later)

stijnvanhulle:
@Ericlm Thanks for opening this pr, I changed some things to align with other plugins and also added some examples for the new Cypress plugin. For now ,this will be a basic(just creating the `cy.request`) implementation, more to follow if there is a need for tasks or more specific commands.

Ericlm:
Thank you very much for you help! And thank you for Kubb in general 😄

PR #1610, opened 2025-03-21
Correct order for null field - Fixes #1584

### LINKED ISSUE #1584: Optional null field results in Zod schema syntax error

3.6.1

Linux

### How does your `kubb.config.ts` config look like

```typescript
import { defineConfig } from '@kubb/core'
import { pluginOas } from '@kubb/plugin-oas'
import { pluginZod } from '@kubb/plugin-zod'

export default defineConfig({
  name: 'Test',
  root: './src',
  input: {
    data: JSON.stringify({
      openapi: '3.1.0',
      components: {
        schemas: {
          TestSchema: {
            properties: {
              stringField: {
                type: 'string',
              },
              nullField: {
                type: 'null',
              },
            },
            type: 'object',
            required: [],
          },
        },
      },
    }),
  },
  output: {
    path: './__generated__',
    clean: true,
  },
  plugins: [pluginOas(), pluginZod()],
})
```

```JSON
{
  "zod": "3.24.2",
  "react": "19.0.0",
  "typescript": "5.7.3"
}
```

Given the example kubb.config.ts file, run the generator and observe the syntax error in the generated zod schema:

```ts
import { z } from "zod";

export const testSchemaSchema = z.object({
      "stringField": z.string().optional(),
  "nullField": .optional()z.null()
      })
```

Every time

The null field should match the behaviour of the string field

```ts
import { z } from "zod";

export const testSchemaSchema = z.object({
    "stringField": z.string().optional(),
    "nullField": z.null().optional()
})
```

### Additional information

_No response_

lambdank:
Hi 👋
I've experienced the bug #1584 as well, and had quick glance at the code.
It seems that `null` has the lowest sort order, and thus always comes last. But this seems to be an issue, since keywords that generates methods should always come after keywords that generate objects. Since `null` produces an object and `optional` produces a method, the current result is `.optional()z.null()`.
The fix seems to be to move `null` above the method keywords.

Feel free to merge this pull request, perform the fix yourselves or reject it if I have misunderstood something.

I really appreciate this project a lot and I'm very grateful for its existence. Feel free to reach out to me, if I can help with anything further.

### DIFF
--- a/packages/plugin-zod/src/parser.ts
+++ b/packages/plugin-zod/src/parser.ts
@@ -155,6 +155,7 @@ export function sort(items?: Schema[]): Schema[] {
     schemaKeywords.password,
     schemaKeywords.matches,
     schemaKeywords.uuid,
+    schemaKeywords.null,
     schemaKeywords.min,
     schemaKeywords.max,
     schemaKeywords.default,
@@ -162,7 +163,6 @@ export function sort(items?: Schema[]): Schema[] {
     schemaKeywords.optional,
     schemaKeywords.nullable,
     schemaKeywords.nullish,
-    schemaKeywords.null,
   ]
 
   if (!items) {


quintal-william:
Thank you for picking this up!! 👏🏼 Lets hope it gets merged soon 😄

stijnvanhulle:
@lambdank Indeed you are right here, `null` should always be last because in the case of a string/number it should be added as last property instead of in the beginning of the array. Thanks for taking this up, this will be added soon as fix for Kubb.

stijnvanhulle:
@all-contributors please add @lambdank  for code

stijnvanhulle (approved):

PR #1636, opened 2025-04-11
feat: support for custom QueryClient

stijnvanhulle:


aburgel:
@stijnvanhulle I'm seeing typescript compiler issues with this change:

```
src/api/gen/hooks/useCreateItem.ts:75:25 - error TS2525: Initializer provides no value for this binding element and the binding element has no default value.

75     mutation: { client: queryClient, ...mutationOptions } = {},
```

looks like if you set a default value of `undefined` for queryClient, it goes away. (i'm on typescript 5.5.4)

PR #1654, opened 2025-04-15
fix(plugin-ts): `nullable: true` now generates `| null` union

### LINKED ISSUE #1653: `additionalProperties` with `nullable: true` do not generate the `| null` union type

v3.9.2

MacOS

### How does your `kubb.config.ts` config look like

```typescript
import { BarrelType, defineConfig, UserConfig } from '@kubb/core';
import { pluginClient } from '@kubb/plugin-client';
import { pluginFaker } from '@kubb/plugin-faker';
import { pluginMsw } from '@kubb/plugin-msw';
import { pluginOas, type Schema } from '@kubb/plugin-oas';
import { pluginReactQuery } from '@kubb/plugin-react-query';
import { pluginZod } from '@kubb/plugin-zod';
import { resolve } from 'node:path';

type CreateUserConfigOptions = {
    apiSpecificationTarget?: string;
    banner?: string;
    clientDataReturnType?: 'full' | 'data';
    clientImportPath: string;
    enableReactQuery?: boolean;
    type: 'web' | 'mobile';
};

type EnumSchema = Extract<Schema, { keyword: 'enum' }>;

const createWebConfig = ({
    apiSpecificationTarget = resolve(__dirname, 'myspec.yml'),
    banner = `/* eslint-disable */
// @ts-nocheck`,
    clientDataReturnType = 'full',
    clientImportPath,
    enableReactQuery = true,
    type,
}: CreateUserConfigOptions): UserConfig => {
    const defaultConfigOptions = {
        banner,
        barrelType: 'all' as BarrelType,
    };

    const plugins: UserConfig['plugins'] = [
        pluginOas(),
        pluginTs({
            output: {
                ...defaultConfigOptions,
                path: './model',
            },
        }),
        pluginClient({
            dataReturnType: clientDataReturnType,
            importPath: clientImportPath,
            output: {
                ...defaultConfigOptions,
                path: './api',
            },
        }),
        pluginFaker({
            mapper: {
                authors: 'faker.helpers.multiple(faker.book.author, { count: 1 })',
            },
            output: {
                ...defaultConfigOptions,
                path: './mocks',
            },
            regexGenerator: 'randexp',
            transformers: {
                name(name, type) {
                    return type === 'function' ? `${name}Mock` : name;
                },
                schema({ name, parentName, schema }, defaultSchemas) {
                    if (schema === undefined) {
                        return undefined;
                    }

                    if ('enum' in schema && 'x-enumNames' in schema) {
                        const updatedSchema = structuredClone(defaultSchemas);
                        const enumEntry: EnumSchema | undefined = updatedSchema.find(
                            (entry): entry is EnumSchema => entry.keyword === 'enum',
                        );

                        if (enumEntry === undefined) {
                            return undefined;
                        }

                        enumEntry.args.items = enumEntry.args.items.map((item) => ({
                            format: item.format,
                            name: typeof item.value === 'string' || typeof item.value === 'number' ? item.value : '',
                            value: item.name,
                        }));

                        if (name === 'firstName' || name === 'lastName') {
                            return [{ keyword: name }];
                        }

                        if (name === 'authors' && parentName === 'Book') {
                            return [{ keyword: 'authors' }];
                        }

                        return updatedSchema;
                    }

                    return undefined;
                },
            },
        }),
        pluginMsw({
            baseURL: '*',
            handlers: false,
            output: {
                ...defaultConfigOptions,
                barrelType: 'propagate',
                path: './msw/index.ts',
            },
            parser: 'faker',
        }),
        pluginZod({
            output: {
                ...defaultConfigOptions,
                path: './zod',
            },
        }),
    ];

    if (enableReactQuery) {
        plugins.push(
            pluginReactQuery({
                client: {
                    dataReturnType: clientDataReturnType,
                    importPath: clientImportPath,
                },
                output: {
                    ...defaultConfigOptions,
                    path: './react-query',
                },
            }),
        );
    }

    return {
        input: {
            path: apiSpecificationTarget,
        },
        name: `myorg-web-api-client`,
        output: {
            barrelType: 'all',
            clean: true,
            path: `./web-api-client`,
        },
        plugins,
        root: './dist/',
    };
};

export default defineConfig([
    createWebConfig({
        clientImportPath: './client',
        type: 'web',
    }),
]);
```

### Swagger/OpenAPI file?

```yaml
openapi: 3.1.0
info:
  title: Example API
  version: 0.0.1

paths:
  /example:
    get:
      summary: Example endpoint
      description: Example
      operationId: getExample
      responses:
        "200":
          description: Default response
          content:
            application/json:
              schema:
                type: object
                properties:
                  data:
                    type: string
                  links:
                    $ref: '#/components/schemas/Links'
                required:
                  - data
                  - links
      tags:
        - Example
components:
  schemas:
    Links:
      title: Links
      type: object
      description: Represents a link structure used for relationships and actions.
      properties:
        self:
          type: string
          description: Link to the current resource.
      additionalProperties:
        type: string
        nullable: true
      required:
        - self
```

```JSON
N/A
```

With the above configuration and yaml file:

1. Run `kubb generate --config ./src/kubb.web.config.ts`
2. Head over to `dist/web-api-client/model/Links.ts`

Notice that its content is:

```ts
/* eslint-disable */
// @ts-nocheck
/**
 * Generated by Kubb (https://kubb.dev/).
 * Do not edit manually.
 */

/**
 * @description Represents the link structure used for relationships and actions.
 */
export type Links = {
  /**
   * @description Link to the current resource.
   * @type string
   */
  self: string
  [key: string]: string
}
```

Every time

It is expected that Links.ts to contain:

```ts
/* eslint-disable */
// @ts-nocheck
/**
 * Generated by Kubb (https://kubb.dev/).
 * Do not edit manually.
 */

/**
 * @description Represents the link structure used for relationships and actions.
 */
export type Links = {
  /**
   * @description Link to the current resource.
   * @type string
   */
  self: string
  [key: string]: string | null
}
```

### Additional information

I've debugged this issue and I've narrowed it down to the `plugin-ts`'s `parser.ts`, more exactly here: https://github.com/kubb-labs/kubb/blob/main/packages/plugin-ts/src/parser.ts#L262-L267

The fact that the additional properties have an entry `{ keyword: 'nullable' }` doesn't matter because:

1. when `parse` is called on the `map()` it will return `undefined` for it due to [this line](https://github.com/kubb-labs/kubb/blob/main/packages/plugin-ts/src/parser.ts#L24).
2. Even if that line was `nullable: () => factory.keywordTypeNodes.null,` it would still not be added, because at the end of the `map()` + `filter(Boolean)` the code is doing  `.at(0)` which only takes the first item on the array.

My proposal would be to:

* Go through all of the items in the array and make a union of all of the types that come out of the `.map()`, but this would imply having the `nullable: () => factory.keywordTypeNodes.null` line...

or

* At least support the nullable by adding an "if nullable is found in the list of items, make a union of the type that came out of the `.at(0)` + the null keyword.

WDYT?

rmachado-studocu:
Closes #1653 

#### Changes

* Detects when `nullable` is used in `additionalProperties` and creates a union type with the resulting of the previously existing code.
* Then passes that to `createIndexSignature` and uses it as the resulting type.


rmachado-studocu:
I've ran the tests locally and one snapshot - the one related to the `nullableAdditionalProperties` schema - is failing, as expected.
I can update the snapshot, of course, but I would like your opinion on whether the snapshot shows the expected result or not.
Once confirmed I'll update the snapshot accordingly 👍🏻

rmachado-studocu:
> I've ran the tests locally and one snapshot - the one related to the `nullableAdditionalProperties` schema - is failing, as expected. I can update the snapshot, of course, but I would like your opinion on whether the snapshot shows the expected result or not. Once confirmed I'll update the snapshot accordingly 👍🏻

Actually disregard that... I'll push the updated snapshot. The diff will show it 🤦🏻

packages/plugin-ts/src/parser.ts
```diff
@@ -259,14 +269,23 @@ export function parse({ current, siblings, name }: SchemaTree, options: ParserOp
...
-      : undefined
+    let additionalProperties = undefined
+
+    if (current.args?.additionalProperties?.length) {
+      additionalProperties = current.args.additionalProperties
+        .map((schema) => parse({ parent: current, name: name, current: schema, siblings }, options))
+        .filter(Boolean)
+        .at(0) as ts.TypeNode
+
+      const isNullable = current.args?.additionalProperties.some((schema) => schema.keyword === schemaKeywords.nullable)
```
stijnvanhulle: You could also use `isKeyword(schema, chemaKeywords.nullable)`, this will also make the type being inferred correctly

stijnvanhulle (approved):

packages/plugin-ts/src/parser.ts
rmachado-studocu: Will see about it! Thanks for the feedback

rmachado-studocu: [85dcff3](https://github.com/kubb-labs/kubb/pull/1654/commits/85dcff3dc6a1ce3465a447f264d988349ce9c28b)

rmachado-studocu: Actually just noticed my IDE's formatter applied a different code style, just re-ran the `format` script, now it's good: [edb7db9](https://github.com/kubb-labs/kubb/pull/1654/commits/edb7db9741321233044a259856e190f4ad38202e)

😉

rmachado-studocu:
Thank you for merging @stijnvanhulle! 

Just a question related to the release process.
I see that there was no changeset defined, and I didn't read the docs on the CONTRIBUTING.md regarding them 🤦🏻 
This means we'll have to wait until a new PR is done for it to happen or should I open a new PR creating the changeset?

Thanks in advance 🙇🏻

stijnvanhulle:
@rmachado-studocu Indeed, there should be a changelog so we could create a new version. I will add one and release a new version for you :)

rmachado-studocu:
Ok I'll close the PR then 👍🏻 Thank you.

PR #1691, opened 2025-04-28
Rewrite schemas with multiple types.

### LINKED ISSUE #1690: Generated typescript types ignore multiple types in OAS 3.1 type array.

3.10.4

Windows

```typescript
import { defineConfig } from "@kubb/core";
import { pluginClient } from "@kubb/plugin-client";
import { pluginOas } from "@kubb/plugin-oas";
import { pluginTs } from "@kubb/plugin-ts";

export default defineConfig(() => {
  return {
    name: "unravelit",
    root: ".",
    input: {
      path: "api/invalid.json",
    },
    output: {
      path: "./src/generated",
      defaultBanner: false,
    },
    plugins: [
      pluginOas(),
      pluginTs({
        output: {
          path: "./types",
        },

        oasType: false,
      }),
      pluginClient({
        output: {
          path: "./client",
        },
        dataReturnType: "full",
        importPath: "../../client",
      }),
    ],
  };
});
```

### Swagger/OpenAPI file?

```bash
{
  "openapi": "3.1.1",
  "info": {
    "title": "Invalid API | v1",
    "version": "1.0.0"
  },
  "paths": {
    "/api/v1/invalid": {
      "get": {
        "tags": ["Invalid"],
        "operationId": "GetInvalid",
        "responses": {
          "400": {
            "description": "Bad Request",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ValidationProblemDetails"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ValidationProblemDetails": {
        "type": "object",
        "properties": {
          "type": {
            "type": ["null", "string"]
          },
          "title": {
            "type": ["null", "string"]
          },
          "status": {
            "pattern": "^-?(?:0|[1-9]\\d*)$",
            "type": ["null", "integer", "string"],
            "format": "int32"
          },
          "detail": {
            "type": ["null", "string"]
          },
          "instance": {
            "type": ["null", "string"]
          },
          "errors": {
            "type": "object",
            "additionalProperties": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          }
        }
      }
    }
  },
  "tags": [
    {
      "name": "Invalid"
    }
  ]
}
```

```JSON
Not using any external packages.
```

When I generate a Typescript client for an OpenAPI 3.1 schema, the generated client does not include all types that were in the properties schema. The example above generates the following code for `ValidationProblemDetails.ts`:

```typescript
export type ValidationProblemDetails = {
  /**
   * @type null | undefined
   */
  type?: null
  /**
   * @type null | undefined
   */
  title?: null
  /**
   * @pattern ^-?(?:0|[1-9]\d*)$
   * @type null,integer,string | undefined, int32
   */
  status?: string
  /**
   * @type null | undefined
   */
  detail?: null
  /**
   * @type null | undefined
   */
  instance?: null
  /**
   * @type object | undefined
   */
  errors?: {
    [key: string]: string[]
  }
}
```

Notice that all properties are defined as `null`, and the `status` property is missing the `number` option.

None

### What is the expected behaviour?

The generated Typescript type should accept all types that were in the type property of the schema, like this:

```typescript
export type ValidationProblemDetails = {
  /**
   * @type null,string | undefined
   */
  type?: string | null
  /**
   * @type null,string | undefined
   */
  title?: string | null
  /**
   * @type null,integer,string | undefined, int32
   */
  status?: (number | string) | null
  /**
   * @type null,string | undefined
   */
  detail?: string | null
  /**
   * @type null,string | undefined
   */
  instance?: string | null
  /**
   * @type object | undefined
   */
  errors?: {
    [key: string]: string[]
  }
}
```

### Additional information

I am creating a PR that rewrites a schema containing multiple not-`null` types as a `oneOf` schema containing every option (except for `null`) for the type property.

maartenvansambeek:
Fixes #1690 by rewriting schemas that have a `type`-array to an `anyOf`-schema.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved generation of TypeScript types for schemas declaring their type as an array, ensuring correct handling of nullable union types and array structures.
	- Enhanced support for OpenAPI 3.1.0 features, including refined parsing of type arrays and format-specific schema handling.


maartenvansambeek:
Hi, first time committer here, let me know if I missed something!

packages/plugin-oas/src/utils/getSchemaFactory.ts
```diff
@@ -31,9 +31,60 @@ type SchemaResult<TWithRef extends boolean = false> =
 export function getSchemaFactory<TWithRef extends boolean = false>(oas: Oas): (schema?: SchemaObject) => SchemaResult<TWithRef> {
   return (schema?: SchemaObject) => {
     const version = isOpenApiV3_1Document(oas.api) ? '3.1' : '3.0'
+    let schemaObject = oas.dereferenceWithRef(schema)
```
stijnvanhulle: We already added support for type array as part of OpenAPI v3.1 but we forgot to add nullable. See https://github.com/kubb-project/kubb/blob/cf9ab066ce42a9ada05c4e364a7d723ae5b68e7b/packages/plugin-oas/src/SchemaGenerator.ts#L1015-L1015.

Could you have a look and add this logic as part of the previous line/file?
```
const [type] = schemaObject.type as Array<OpenAPIV3.NonArraySchemaObjectType>
```

SOmething like this:
```
 if (schemaObject.type) {
      if (Array.isArray(schemaObject.type)) {
        // OPENAPI v3.1.0: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0
        const items = schemaObject.type.filter((value) => value !== 'null') as Array<OpenAPIV3.NonArraySchemaObjectType>

        const isNullable = schemaObject.type.includes('null')

        return [
          ...items.map((item) =>
            this.parse({
              schemaObject: {
                ...schemaObject,
                type: item,
              },
              name,
              parentName,
            }),
          ),
          ...this.parse({
            schemaObject: {
              ...schemaObject,
              nullable: isNullable,
            },
            name,
            parentName,
          }),
          ...baseItems,
        ].filter(Boolean)
      }
```

maartenvansambeek: Thanks for the quick response! The thing is, there are two problems here. The nullable, as you said, but it is also possible to have multiple types in the type-property, effectively creating a union.

This is why I cheated and converted the type array to union. I could try and take the code from the 'anyOf' branch (line 592) to produce a union, would that be better?

maartenvansambeek:
So, I made the changes in the `SchemaGenerator`, but I had to move some stuff around. 

The initial check using isArray now has the logic for parsing the types, and I had to add some type checks to the `format` and `pattern` properties.

stijnvanhulle:
@all-contributors please add @maartenvansambeek for code

stijnvanhulle:
@maartenvansambeek Thanks for opening a pr and fixing the issue, really appreciate your effort and contribution. To be honest, I was not aware of the possibility to use the updated `type` as an array with more than just nullable.

stijnvanhulle (approved):

maartenvansambeek:
I see the CodeQL is failing, what can I do about that?

stijnvanhulle:
> I see the CodeQL is failing, what can I do about that?

I'm having a look and fixing the Github action for this.

PR #1692, opened 2025-04-28
fix(plugin-faker): Fix types of enums nested in array

aburgel:
With [this PR](https://github.com/kubb-labs/kubb/pull/1648), the faker plugin generates invalid types for enums that are nested in an array.

In my app, it produced something like:

```typescript
export function createThing(data?: Partial<Thing>): Thing {
  return {
    ...{
      attributes: faker.helpers.multiple(() => ({
        enum: faker.helpers.arrayElement<NonNullable<Thing>["enum"]>(["a", "b"]),
      })),
    },
    ...(data || {}),
  };
}
```

it should have been more like:

```typescript
export function createThing(data?: Partial<Thing>): Thing {
  return {
    ...{
      attributes: faker.helpers.multiple(() => ({
        enum: faker.helpers.arrayElement<NonNullable<Thing["attributes"][number]>["enum"]>(["a", "b"]),
      })),
    },
    ...(data || {}),
  };
}
```

that would have correctly extracted the type from the array attribute. The difference being the `Thing["enum"]` does not exist, you have to first access the array attribute (`["attributes"]`) and then extract the array item type (`[number]`).

This PR addresses this, but I have only tested it for one level of depth, and only for arrays. I think something similar may need to be implemented for nested objects.

The schema of my object is:

```yaml
    Thing:
      type: object
      properties:
        attributes:
          type: array
          items:
            type: object
            properties:
              enum:
                type: string
                enum:
                  - a
                  - b
            required:
              - enum
      required:
        - attributes
```


packages/plugin-faker/src/parser.ts
```diff
@@ -194,7 +194,11 @@ export function parse({ current, parent, name, siblings }: SchemaTree, options:
 
   if (isKeyword(current, schemaKeywords.array)) {
     return fakerKeywordMapper.array(
-      current.args.items.map((schema) => parse({ parent: current, current: schema, siblings }, { ...options, canOverride: false })).filter(Boolean),
+      current.args.items
+        .map((schema) =>
+          parse({ parent: current, current: schema, siblings }, { ...options, typeName: `${options.typeName}["${name}"][number]`, canOverride: false }),
```
aburgel, line 199: This is the functional change. I'm constructing a new type based on the current attribute name (which gets you the array) and then using `[number]` to pull out the array item type.

Issue #1699, opened 2025-05-01
Query parameters should always be Zod coerced, because query parameters are always strings


SCdF:
### What version of `kubb` is running?

3.10.5

### What kind of platform do you use?

None

### Your `kubb.config.ts` config file?

```typescript

```

### Swagger/OpenAPI file?

```bash

```

### What version of external packages are you using(`@tanstack-query`, `MSW`, `React`, `Vue`, ...)

```JSON

```

### What steps can reproduce the bug?

Zubb's Zod plugin allows you to define whether or not you coerce types in Zod before attempting to validate them.

However I think that query parameters in the URL are a special case, because they are not json from a body, they should always be considered strings initially. Which means the Zod code should **always** try to coerce them first.

### How often does this bug happen?

None

SCdF:
Please take a look at the PR linked above, I'm fairly sure that fixes this issue.

PR #1709, opened 2025-05-07
Always Coerce Parameters

SCdF:
path and query parameters are part of the url, and so will always be strings.

resolves kubb-labs/kubb#1699

## Summary by CodeRabbit

- **New Features**
  - Enhanced input handling for the "offset" query parameter, allowing string values representing numbers to be automatically converted and accepted.

- **Refactor**
  - Improved internal handling of coercion settings for parameter schemas to better distinguish between global and local coercion behavior.


stijnvanhulle (approved):

SCdF:
@stijnvanhulle I've added a patch note :-) Note that I consider this a patch level change and not a major or a minor, because I think it's a bugfix, and I can't work out how it would be a breaking change for anyone, as it's just adding coercion. If you think it's bigger than that we can change

PR #1724, opened 2025-05-16
Query object should not be optional with defaulted query parameters

### LINKED ISSUE #1698: Query parameter object with all parameters defaulting incorrectly marked as optional in Zod

3.10.5

MacOS

```typescript
import { defineConfig } from "@kubb/core";
import { pluginOas } from "@kubb/plugin-oas";
import { pluginZod } from "@kubb/plugin-zod";
import { pluginTs } from "@kubb/plugin-ts";

export default defineConfig(() => {
  return {
    root: ".",
    input: {
      path: "./openapi.yaml",
    },
    output: {
      path: "./src/_generated/",
    },
    plugins: [
      pluginOas(),
      pluginTs(),
      pluginZod({
        output: {
          path: "zod-only",
        },
      }),
      pluginZod({
        typed: true,
        output: {
          path: "zod-typed",
        },
      }),
    ],
  };
});
```

### Swagger/OpenAPI file?

```bash
openapi: 3.0.0
info:
  title: Kubb Required Default example
  description: I think that this datatype should not end up being optional in Zod
  version: 0.0.1

paths:
  /thing:
    summary: Represents a thing
    description: |
      This resource represents a thing in the system.
    get:
      operationId: getThings
      summary: Get a list of things

      parameters:
        - in: query
          name: limit
          description: Maximum number of things to return
          schema:
            type: integer
            minimum: 1
            maximum: 100
            default: 100
        - in: query
          name: skip
          description: Number of things to skip
          schema:
            type: integer
            minimum: 0
            default: 0

      responses:
        "200":
          description: A list of things
          content:
            application/json:
              schema:
                type: object
                properties:
                  items:
                    type: array
                    items:
                      $ref: "#/components/schemas/Thing"
                required:
                  - items

components:
  schemas:
    Thing:
      type: object
      properties:
        limit:
          description: Maximum number of things to return
          type: integer
          minimum: 1
          maximum: 100
          default: 100
        skip:
          description: Number of things to skip
          type: integer
          minimum: 0
          default: 0
```

```JSON
{
  "name": "kubb-required-default",
  "version": "1.0.0",
  "description": "",
  "type": "module",
  "scripts": {
    "start": "node --experimental-strip-types ./src/index.ts",
    "generate": "kubb generate"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "packageManager": "pnpm@8.13.1+sha512.758bebc6989cb6462c5138129840193b10c258ddcdb0ab43ee0ef9dd87784bc218d6e41cb5beeac362b8ce33f1f8b08025b599933c0bce82050a9a36b8474d43",
  "dependencies": {
    "@kubb/cli": "^3.10.5",
    "@kubb/core": "^3.10.5",
    "@kubb/plugin-oas": "^3.10.5",
    "@kubb/plugin-ts": "^3.10.5",
    "@kubb/plugin-zod": "^3.10.5"
  }
}
```

I have created a test repo, please see: https://github.com/SCdF/kubb-required-default/blob/main/src/index.ts

This repo walks through two bugs, though I'm not sure about the second one, as I feel it's probably more philosophical.

The main issue is the first one, which is that when all of your query parameters have defaults the Zod code correctly considers each parameter as required (because if you don't pass a value it will default one), but it still types the entire query parameter object as optional (`| undefined`).

Every time

### What is the expected behaviour?

The Zod code should return a non-optional type, because no matter what you pass it you will always get an object, because of the defaulting.

### Additional information

This looks semi related to https://github.com/kubb-labs/kubb/issues/1290

I also don't think the Typescript plugin is right either (see the second half of my example in the repo), but that feels more like a philosophical disagreement than an actual bug.

SCdF:
Changes Zod logic to no longer consider a Query parameter Zod object optional if at least one of the parameters has a default value. 

This means if you default at least one value, the resulting query parameter object from Zod will, typescript-wise, be guaranteed to exist.

Fixes #1698 

## Summary by CodeRabbit

- **Bug Fixes**
  - Query parameter objects are now required if any parameter within them has a default value, enforcing stricter validation for API requests.

- **New Features**
  - Added new validation schemas and test cases to ensure correct handling of required query parameters with default values.

- **Tests**
  - Introduced new test cases to verify the updated query parameter validation logic.


SCdF:
I don't know why the typecheck check is failing, it seems unrelated to my changes and fails in upstream/main for me as well. @stijnvanhulle let me know if I've done anything wrong, but afaict this is a global issue.

PR #1733, opened 2025-06-01
Fix: Cannot mutate a Store directly

hugoFelippe:
Hi, I'm using `@kubb/plugin-solid-query` and I'm getting the warning: **"Cannot mutate a Store directly"**.
I looked into it and found [this line in the source code](https://github.com/kubb-labs/kubb/blob/6de53397f285f7c2cc9cbe2dd86d9b4569820a28/packages/plugin-solid-query/src/components/Query.tsx#L175C1-L175C46), which seems problematic.

Since Solid's `createStore` doesn't allow direct mutation (you have to use the setter), this line might not be working as intended.

To fix i removed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue causing the error "Cannot mutate a Store directly" when using query features.

- **Documentation**
  - Updated the changelog to document the recent bug fix.


stijnvanhulle:
@all-contributors please add @hugoFelippe for code

PR #1805, opened 2025-07-15
fix: update argument types for vue-query; update argument unwrapping

th3l0g4n:
This will update the types for vue-query so it is possible to pass getters as property keys.

For example (pseudo code):
```
const queryReturn = useGetPetDetails({ id: () => pet.id })
```
Whereas pet is a reactive.
The usage above is at the moment not possible. Using plain pet.id would currently never update.

## Summary by CodeRabbit

* **New Features**
  * Functions now support both Vue refs and getter functions as input parameters, offering greater flexibility when passing reactive or computed values.

* **Refactor**
  * Updated type annotations and internal value extraction to support a broader range of input types, ensuring compatibility with both refs and getter functions.
  * Public function signatures have been updated to reflect the new supported input types.

* **Style**
  * Improved consistency in handling reactive and computed values throughout the codebase.

### DIFF
--- a/.changeset/sweet-jars-itch.md
+++ b/.changeset/sweet-jars-itch.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-vue-query": patch
+---
+
+use MaybeRefOrGetter as argument type; use toValue for unwrapping arguments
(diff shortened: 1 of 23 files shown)


stijnvanhulle (approved):

stijnvanhulle:
@all-contributors please add @th3l0g4n for code

th3l0g4n:
Thx for reviewing this so quickly. This Change is currently really needed  :)

PR #1829, opened 2025-08-05
fix(parser-ts): update prettier to be a peer dependency

pkasarda:
## Summary by CodeRabbit

* **Chores**
  * Updated dependency management for Prettier, moving it to development and peer dependencies.

## Nitpick from CodeRabbit

75-77: Consider clarifying whether the peer dependency is optional
If @kubb/parser-ts can be used without runtime access to prettier (e.g., only when specific APIs are invoked), add:
```json
"peerDependenciesMeta": {
  "prettier": {
    "optional": true
  }
}
```
Marking it optional prevents npm from emitting warnings for consumers that don’t need Prettier, while still expressing the expected major range.

### DIFF
--- a/packages/parser-ts/package.json
+++ b/packages/parser-ts/package.json
@@ -63,14 +63,17 @@
     "typecheck": "tsc -p ./tsconfig.json --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
-    "prettier": "^3.6.2",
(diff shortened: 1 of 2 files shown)


stijnvanhulle:
Replaced with https://github.com/kubb-labs/kubb/pull/1847

Issue #1864, opened 2025-08-28
Infinite query type issues


filipjakov:
### What version of `kubb` is running?

3.18.2

### What kind of platform do you use?

MacOS

### Your `kubb.config.ts` config file?

```typescript
not relevant
```

### Swagger/OpenAPI file?

```bash

```

### What version of external packages are you using(`@tanstack-query`, `MSW`, `React`, `Vue`, ...)

```JSON

```

### What steps can reproduce the bug?

I have 2 issues with the generated types:

1. infinite query `page` param type error. Example:

```tsx
export function officeControllerInfiniteQueryOptions(
	params: OfficeControllerQueryParams,
	config: Partial<RequestConfig> & { client?: typeof fetch } = {}
) {
	const queryKey = officeControllerInfiniteQueryKey(params);
	return infiniteQueryOptions<
		ResponseConfig<OfficeControllerQueryResponse>,
		ResponseErrorConfig<OfficeController500>,
		ResponseConfig<OfficeControllerQueryResponse>,
		typeof queryKey,
		number
	>({
		enabled: !!params,
		queryKey,
		queryFn: async ({ signal, pageParam }) => {
			config.signal = signal;

			if (params) {
				params['page'] =
					pageParam as unknown as OfficeControllerQueryParams['page']; // <-- ERRORS
			}
			return officeController(params, config);
		},
		initialPageParam: 1,
		getNextPageParam: (lastPage, _allPages, lastPageParam) =>
			Array.isArray(lastPage.data) && lastPage.data.length === 0 ? undefined : lastPageParam + 1,
		getPreviousPageParam: (_firstPage, _allPages, firstPageParam) =>
			firstPageParam <= 1 ? undefined : firstPageParam - 1,
	});
}
```

where i get the following error:

```tsx
Type 'number | undefined' is not assignable to type 'number'.
  Type 'undefined' is not assignable to type 'number'.ts(2322)
```

for the line denoted with ERRORS above. 

2. The generated infinite query hook contains a generic in the function definition which is not being used, erroring in TS as well.

```tsx
export function useOfficeControllerInfinite<
	TData = InfiniteData<ResponseConfig<OfficeControllerQueryResponse>>,
	TQueryData = ResponseConfig<OfficeControllerQueryResponse>, // <-- ERROR
	TQueryKey extends QueryKey = OfficeControllerInfiniteQueryKey,
>(
```

I get the following error:

```tsx
[...]

PR #1890, opened 2025-09-20
fix(plugin-react-query): align infinite query generics with tanstack

humarkx:
## Summary

  Official builds of `@kubb/plugin-react-query` still emit infinite-query hooks that target the old TanStack overloads where
  `TQueryData` doubles as `TPageParam`. With TanStack Query v5 (and the current `@tanstack/react-query` types), those signatures now require a distinct `TPageParam`, so consumers hit `TS2769` when `pageParam` isn’t a `number`.

  This change updates the plugin to:

  - Emit `InfiniteQuery` hooks with explicit `TQueryFnData`, `TError`, `TData`, `TQueryKey`, and `TPageParam` generics (defaulting
  `TPageParam` to the configured query param type or `number`).
  - Mirror the same generics in `InfiniteQueryOptions`, so the builder returns options keyed by the correct page-param type.
  - Pass the configured `queryParam` from the generator into `InfiniteQuery`, letting the emitter pick the right cursor/offset type.

  Generated consumers (tested in our project against TanStack Query v5.89.0) typecheck without custom patching.

### DIFF
--- a/packages/plugin-react-query/src/components/InfiniteQuery.tsx
+++ b/packages/plugin-react-query/src/components/InfiniteQuery.tsx
@@ -3,7 +3,7 @@ import type { OperationSchemas } from '@kubb/plugin-oas'
 import { getComments, getPathParams } from '@kubb/plugin-oas/utils'
 import { File, Function, FunctionParams } from '@kubb/react'
 import type { ReactNode } from 'react'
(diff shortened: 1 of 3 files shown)


humarkx:
Fixes #1864

stijnvanhulle:
@all-contributors please add @humarkx for code

Issue #1908, opened 2025-09-24
Generated `operations.ts` exceed Typescript serialization limit


Jakub-Cerovsky:
### What version of `kubb` is running?

3.16.1

### What kind of platform do you use?

Windows

### Your `kubb.config.ts` config file?

```typescript
pluginZod({
      output: {
        path: './schemas',
        barrelType: 'named',
      },
      dateType: 'stringOffset',
      unknownType: 'unknown',
      operations: true,
      exclude: [{ type: 'method', pattern: 'options' }],
      coercion: {
        strings: false,
        numbers: true,
        dates: false,
      },
    }),
```

### Swagger/OpenAPI file?

```bash
Large (~80 endpoints)
```

### What version of external packages are you using(`@tanstack-query`, `MSW`, `React`, `Vue`, ...)

```JSON

```

### What steps can reproduce the bug?

1. Large OpenAPI file.
2. `tsconfig.json` with `declaration: true`
3. `plugin-zod`
4. `kubb generate`

### How often does this bug happen?

Every time

### What is the expected behaviour?

Everything was working fine until now, but today, I have introduced 2 new endpoints which, caused the typescript serializer to go over the limit. 

<img width="1589" height="130" alt="Image" src="https://github.com/user-attachments/assets/7ea7c1bd-cb31-4595-9f7b-d944260c8df3" />

I assume, the operations needs to be split into groups similar to schemas? This problem would not happen if I would set tsconfig declarations to false, but I am using these generated files in multiple packages and hence, need those.

Same thing happens for paths.

stijnvanhulle:
@Jakub-Cerovsky Fixed in Kubb v4.1.2 :) We removed the `as const` so TypeScript has less issues when using a big const

ghost:
@stijnvanhulle Thank you for a quick response, however, the fix you have made is around `plugin-client`, the issue I was addressing is for `plugin-zod`. 😅

I would suggest making a change of asserting a type to the `operations` object, something of this sort in https://github.com/kubb-labs/kubb/blob/main/packages/plugin-zod/src/components/Operations.tsx:

```
<Const export name={name} type="Record<string, SchemaNames>">
  {`{${transformers.stringifyObject(operationsJSON)}}`}
</Const>
```

stijnvanhulle:
@Jakub-Cerovsky Your changes have been added to Kubb v4.14

PR #1916, opened 2025-10-07
fix: Add type for operations generated by `plugin-zod`

ghost:
Hello 👋

First of all, thank you for this amazing project!

I have previously opened an issue (#1908) related to generated `operations.ts` file, which "crashes" once too large. The issue got closed as solve, however, that only addressed the `plugin-client`. My issue persists, and after testing, removing `asConst` would not resolve it.

image

This fix introduces a `OperationType` and `OperationsMap` both of which should match the generated structure of operations and help Typescript with object serialization overflow (tested and works).

## Summary by CodeRabbit

- **New Features**
  - Added typed operation schemas and a global operations map to provide stronger typing for requests, parameters, responses, and errors.
  - Generator now inserts a type-only import from a configurable path, improving integration with external type libraries.
  - Exposed a new configuration option to customize the import path for those type imports.

- **Chores**
  - Added a changeset marking a major release with a breaking change to operation typings.


ghost:
@stijnvanhulle I think the PR is ready for a final review :)

ghost:
Quick update on this — after some extra testing and digging around, I think I found a better solution.  

The original issue happens because the current package builds with **inline `zod` imports**, which makes TypeScript throw a **serialization error** when dealing with large objects.

This PR fixed that by replacing the inline `zod` types with `any`, which got rid of the error but also meant we lost the generated `zod` types.

Turns out, just adding:

```ts
import type { z } from 'zod';
```
…and letting the build use that instead of inlining everything keeps the zod types intact and fixes the serialization issue.

**Current built file**
image

**File built with imported zod**
image

**Pros**
- Keeps the generated zod types (no more any).
- No more TypeScript serialization errors.

**Cons**
- Requires leaving an unused `import type { z } from 'zod';` in the generated file just to make TypeScript happy.

Overall, this seems like a much cleaner solution than the first one — minimal change, keeps type safety. What do you think? The only change needed would be the one currently in the [operationsGenerator.tsx](https://github.com/kubb-labs/kubb/pull/1916/files#diff-57608956cbc6989db7b82e409a3cc274cb507835baaf96a619d065608e2c7dd2). If you agree, I will create a different PR just with this change, so it is more clean and understandable.

stijnvanhulle:
@all-contributors please add @Jakub-Cerovsky for code

stijnvanhulle:
@Jakub-Cerovsky thanks a lot on opening a pr and contributing to Kubb. I appreciate that a lot! 

Looks good to me, this can be added to the next version of Kubb :) 

Another question: what are your thoughts about using jsx for creating files? Easy enough, are we missing some features to make it as easy as possible, ...?

PR #2004, opened 2025-11-07
feat: typing MSW callbacks

Ericlm:
This PR adds support for typing of path params, request and responses for custom MSW callbacks.

However, I don't know which type the handlers should accept by default; I guess the params and requests are the same, but the response will be different.


Ericlm:
I guess the type should be something like `ResponseResolver<Params, Body, Response400 | Response200>` ?

What do you think @stijnvanhulle?

PR #2075, opened 2025-11-27
Fix/build form data json contenttype

GKNewsrooms:
Fix buildFormData to correctly include Content-Type: application/json

### DIFF
--- a/.changeset/fix-buildformdata-content-type.md
+++ b/.changeset/fix-buildformdata-content-type.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-client": patch
+---
+
+Fix buildFormData to correctly include Content-Type: application/json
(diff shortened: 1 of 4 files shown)


GKNewsrooms:
Presumably this needs to regenerate examples/** since the config gets copied there but I couldnt get it to work. if one of the maintainers wants to help me please feel free

GKNewsrooms:
Reasons why I did this: basically all web servers will 415 if you send json via multipart with the wrong content-type

stijnvanhulle:
@copilot

> Presumably this needs to regenerate examples/** since the config gets copied there but I couldnt get it to work. if one of the maintainers wants to help me please feel free

I will update this after merging so the examples are up to date

stijnvanhulle:
@all-contributors please add @GKNewsrooms for code

PR #2154, opened 2025-12-10
Allow overriding faker data when schema uses `allOf`

IanVS:
I hit an issue with Kubb where my schemas that combined object schemas using `allOf` had an unused `data` parameter, which was not being used for override.  See https://github.com/kubb-labs/kubb/pull/2154/commits/ac1dc0dc0b6aaaa94b2e597bac232cbebb33cc89 as an example of what would be generated before the other changes being made in this PR.

I traced it down to the use of `Object.assign`, which wasn't treated as a type that could be overridden. 

I think there are two ways this could be solved:

1. Check for `Object.assign(` in addition to `{`:

```diff
diff --git a/packages/plugin-faker/src/components/Faker.tsx b/packages/plugin-faker/src/components/Faker.tsx
index f9e6c38ad..bf7fc6db4 100644
--- a/packages/plugin-faker/src/components/Faker.tsx
+++ b/packages/plugin-faker/src/components/Faker.tsx
@@ -36,7 +36,7 @@ export function Faker({ tree, description, name, typeName, seed, regexGenerator,
   )

   const isArray = fakerText.startsWith('faker.helpers.arrayElements') || fakerText.startsWith('faker.helpers.multiple')
-  const isObject = fakerText.startsWith('{')
+  const isObject = fakerText.startsWith('{') || fakerText.startsWith('Object.assign(');
   const isTuple = fakerText.startsWith('faker.helpers.arrayElement')

   const isSimpleString = name === 'string'
```

2.  Update the parser to use spread object notation instead of `Object.assign`.  

This has a side benefit of generating code that will fail to parse when invalid inputs are provided to `allOf`.  For example, in this package's own mock data, there was/is string & number fields being provided to `and`, which is impossible.  Something can't be both a string and a number.  Only objects are intended to be provided to `allOf`.  With approach 1 above, kubb would continue to generate unexpected, but not necessarily invalid data.  For example:

```ts
Object.assign({}, "abc", 123);  // -> { 0: "a", 1: "b", 2: "c" }
```

So, I took approach 2 here.  I welcome any feedback or other ideas you might have.

### DIFF
--- a/.changeset/lucky-doodles-invent.md
+++ b/.changeset/lucky-doodles-invent.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-faker": patch
+---
+
+Allow overriding data in mocks when using allOf schemas


packages/plugin-faker/src/__snapshots__/parser.test.ts.snap
```diff
@@ -52,7 +52,7 @@ exports[`faker parse > 'number' 1`] = `"faker.number.float()"`;
 
 exports[`faker parse > 'object' 1`] = `"{"firstName": faker.string.alpha({ length: 2 }),"address": faker.string.alpha()}"`;
 
-exports[`faker parse > 'objectAnd' 1`] = `"{"firstName": faker.string.alpha({ length: 2 }),"age": faker.number.float({ min: 3 }),"address": Object.assign({}, faker.string.alpha(), faker.number.float())}"`;
+exports[`faker parse > 'objectAnd' 1`] = `"{"firstName": faker.string.alpha({ length: 2 }),"age": faker.number.float({ min: 3 }),"address": {...faker.string.alpha(), ...faker.number.float()}}"`;
```
IanVS: Note, I did not update this part of the schema, which is an address field that is both a string and a number (impossible).  I wasn't certain what the intent was of `objectAnd`, so I left it alone.

stijnvanhulle: @IanVS You can merge main back into your branch, I updated the test cases to be more real life friendly

stijnvanhulle:
@IanVS looks good to me, thanks for having a look. Could you just add a changelog with pnpm changeset?

packages/plugin-faker/src/__snapshots__/parser.test.ts.snap
IanVS: Done!

PR #2170, opened 2025-12-15
Cypress: add "prefix" option for generated requests

Ericlm:
Add a new `prefix` option in the Cypress plugin, to modify function's name.
```ts
// The configuration
{
    prefix: 'prefix'
},

// Generates
export function prefixDeletePetsPetid(): Cypress.Chainable<DeletePetsPetidMutationResponse> {
  return cy.request('delete', '/pets/:petId', undefined).then((res: Cypress.Response<DeletePetsPetidMutationResponse>) => res.body)
}
```

It avoids conflicts between the name of function generated for the plugin-client VS plugin-cypress.

It also set the default prefix to `cy` to "namespace" it by default.

### DIFF
--- a/packages/plugin-cypress/src/components/Request.tsx
+++ b/packages/plugin-cypress/src/components/Request.tsx
@@ -1,10 +1,10 @@
+import { camelCase } from '@kubb/core/transformers'
 import { URLPath } from '@kubb/core/utils'
 import { type HttpMethod, isOptional } from '@kubb/oas'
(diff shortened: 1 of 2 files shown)


stijnvanhulle (changes requested):
@Ericlm adding default as 'cy' will be a breaking change, does the transformers.name not work for cypress? That one can change the names of functions.

Ericlm:
@stijnvanhulle I used the transformers, and it works perfectly; sorry for making this PR 😅

PR #2179, opened 2025-12-17
Feature: add options field to cypress functions

Ericlm:
Add an option parameter in the generated Cypress requests corresponding to the cy.request options param. It allows headers etc. and also overriding existing one if needed.

*Before:*
```ts
export function createPets(data: CreatePetsMutationRequest): Cypress.Chainable<CreatePetsMutationResponse> {
  return cy.request('post', '/pets', data).then((res: Cypress.Response<CreatePetsMutationResponse>) => res.body)
}
```

*After:*
```ts
export function createPets(data: CreatePetsMutationRequest, options?: Partial<Cypress.RequestOptions>): Cypress.Chainable<CreatePetsMutationResponse> {
  return cy
    .request({
      method: 'post',
      url: '/pets',
      body: data,
      ...options,
    })
    .then((res: Cypress.Response<CreatePetsMutationResponse>) => res.body)
}
```

### DIFF
--- a/.changeset/open-birds-like.md
+++ b/.changeset/open-birds-like.md
@@ -1,0 +1,5 @@
+---
+"@kubb/plugin-cypress": patch
+---
+
+Add an options param to generated requests, accepting the same options as https://docs.cypress.io/api/commands/request#Arguments


Ericlm:
Thank you! I'll merge this as soon as all the checks have passed.

Issue #2184, opened 2025-12-17
Support defining custom hook options for generated React Query hooks


JoaoBrlt:
### What is the problem this feature would solve?

Currently the generated React Query hooks are great but in order to use them in our apps, we often need to define additional options like `onSuccess`, `onError`, `meta`, etc. in order to invalidate other queries, show an error flag when a mutation fails, show a success flag when a mutation succeeds, configure the error message, configure the success message, etc.
When using the generated React Query hooks in multiple apps, this leads to a lot of code duplication and maintenance overhead.
For example, if we forget to invalidate a query, it can easily result in stale UI state which is not great.
Today, we partially address that by creating and maintaining wrapper hooks to add the custom options.
However, we think this defeats a bit the purpose of generating React Query hooks.

### External documents/projects?

I've opened a discussion [here](https://github.com/kubb-labs/kubb/discussions/2171) to discuss a little bit about the cache invalidation use case.
Indeed, I've discovered that there was an example in Kubb v2 using templates to generate an `invalidations.ts` file and update the generated React Query mutation hooks to use a manually defined `useInvalidationForMutation()` hook (see https://github.com/kubb-labs/kubb/pull/853).
That's exactly the kind of solution I was looking for because it avoids the wrapper hooks and code duplication, it is very flexible (the declaration of queries to invalidate is optional) and type-safe (thanks to the generated `Invalidations` type).

However, when doing some research, I discovered that [Orval](https://orval.dev/) has 2 options ([queryOptions](https://orval.dev/reference/configuration/output#queryoptions) and [mutationOptions](https://orval.dev/reference/configuration/output#mutationoptions)) to define custom query options and mutation options by providing the path to a JS file.
The only thing they are missing is the type-safety (which is quite important in my opinion) and they don't provide the hook name as a parameter (which means we have to depend on the mutation key for example).
I think that's a good idea because it offers developers even more flexibility to define the custom options to be used by the generated React Query hooks (if we can solve the type-safety issue).

### What is the feature you are proposing to solve the problem?

Introduce a new `customOptions` option to the `@kubb/plugin-react-query` plugin. 
The option would take the path to a JS file (similar to how the `client` option works today) and the name of the exported hook.
When the option is set, the plugin would generate a type (for example, `HookOptions`) which maps hook names to hook options (like the `invalidations.ts` file in the example).
[...]

PR #2185, opened 2025-12-17
Feature: Add support for custom hook options

JoaoBrlt:
Related issue: https://github.com/kubb-labs/kubb/issues/2184

Changes:
- Add a `customOptions` option to the `@kubb/plugin-react-query` plugin.
- Add an `output.override` option to all plugins to control whether to override existing external files.
- The `customOptions` option takes:
  - `importPath`: the path to the JS file defining the hook to retrieve the custom hook options.
  - `name`: the name of the exported hook.
- When the `customOptions` option is set:
  - Generate a `HookOptions` type which maps the hook options to each hook name.
  - Update the code of the generated React Query hooks to use the custom hook to retrieve the custom hook options.
  - If the custom options hook file does not exist, a default one is generated.
- Update the `react-query` example to use custom options.
- Update the generator tests to test the custom options.
- Make the `OperationGenerator.getOptions` method public in order to be able to get the options of an operation.

This is a first draft of the feature. 
I'm really open to feedback and would be happy to solve the use case differently.
Moreover, it's my first contribution to the project so don't hesitate to tell me if there's something I'm missing.


packages/plugin-react-query/src/types.ts
```diff
@@ -85,6 +85,20 @@ export type Infinite = {
   initialPageParam: unknown
 }
 
+type CustomOptions = {
+  /**
+   * Path to the hook that will be used to customize the hook options.
+   * It will be used as `import ${customOption.name} from '${customOptions.importPath}'`.
+   * It allows both relative and absolute paths but be aware that we will not change the path.
+   */
+  importPath: string
```
stijnvanhulle, line 94: Should we create a default generated hooks for this?

stijnvanhulle:
@JoaoBrlt would it make sense to use `operationdId` instead of the hook name?

packages/plugin-react-query/src/types.ts
JoaoBrlt: Ah, that's a good idea. :ok_hand: It would simplify the set up process. We can probably generate a default file if the file referred in the `importPath` does not already exist. Is that what you were thinking too?

stijnvanhulle: @JoaoBrlt Yes something like that, the only thing that will not be easy is what if we need to change that generated hook. If the user already created that file, Kubb need to have like a confirmation action to override or ignore. Just thinking out loud here.

Something like this could also be used for `client.importPath`

JoaoBrlt:
> @JoaoBrlt would it make sense to use `operationdId` instead of the hook name?

@stijnvanhulle I think using only the `operationId` would not work in some cases. The first example that comes to my mind is an operation like `GET /pet/findByTags` that produces both a query hook and an infinite query hook. In this case, if I remember correctly the hook options and return types are slightly different so using the generated `HookOptions` type would probably not be possible. Moreover, I can see use cases where we would like to define some custom hook options for the query hook and not the infinite query hook (for example, for the stale time).

However, maybe we can provide the `operationId` as an additional parameter to the `useCustomHookOptions` hook as a context if developers want to define custom hook options by `operationId` without the type-safety (which might also be a use case)?

packages/plugin-react-query/src/types.ts
JoaoBrlt: Hmm, indeed, maybe we can solve that by simply adding an optional property to `customOptions` (for example, `overrideExistingFile`) which defaults to `false`?
This way, by default, if the specified file does not exist, we generate a default hook.
Otherwise, if the specified file already exists, we do nothing (to preserve existing custom hook options).
All that while still offering a way to resynchronize the file with a fresh example when upgrading Kubb for example (by setting `customOptions.overrideExistingFile` to `true`).

stijnvanhulle: Or a more generic approach: `output.override` that when set will override the `importPath`. This means we could also support other overrides like for the client.importPath and more in the future. Thoughts?

JoaoBrlt: Indeed, this is a good idea, especially if other file overrides are supported later on.

JoaoBrlt: I won't have time today, but I'll take a look at it tomorrow. :wink: Unless you already have an implementation in mind. Don't hesitate to update the PR as you see fit.

JoaoBrlt: Hey @stijnvanhulle :wave: I've pushed an implementation to generate the custom hook options file if it does not already exist and support a new `output.override` option to override the file when required. 

For now, the `output.override` option only controls the override of the custom hook options file but we can use it later for the client file as well. One thing I don't know is whether we should set the `output.override` option of all the plugins when setting the option in the root of the `kubb.config.ts` file? I didn't find how to do that. 

Let me know what you think. :wink:

packages/core/src/types.ts
```diff
@@ -328,6 +328,11 @@ export type Output<TOptions> = {
    * Add a footer text in the beginning of every file
    */
   footer?: string | ((options: TOptions) => string)
+  /**
+   * Whether to override existing external files if they already exist.
```
stijnvanhulle, line 341: We also have `output` in every plugin so maybe we could move it to that place instead. This makes sure we can set override per plugin instead for all generated files/plugins.

packages/plugin-react-query/src/generators/hookOptionsGenerator.tsx
```diff
@@ -0,0 +1,196 @@
...
+        }
+        if (isMutation(operation)) {
+          acc[getHookName(operation)] = `Partial<ReturnType<typeof ${getMutationHookOptions(operation)}>>`
+        }
+        return acc
+      },
+      {} as Record<string, string>,
+    )
+
+    if (!options.customOptions) {
```
stijnvanhulle: Could we move this higher so that Kubb does not need to call all resolves like `getName(operation, { type: 'function', prefix: 'use', suffix: 'infinite' })` before seeing that we don't need it?

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
```diff
@@ -0,0 +1,65 @@
...
+      path: path.resolve(hookIndexFile.path, importPath),
+    }
+
+    if (fs.existsSync(file.path) && !override) {
+      return null
+    }
+
+    return (
+      <File baseName={file.baseName} path={file.path}>
+        <File.Import name={['QueryClient']} path="@tanstack/react-query" isTypeOnly />
```
stijnvanhulle: We support an `importPath` option to override the default tanstack react-query import so best to replace path here by 

```
    const importPath = options.query ? options.query.importPath : '@tanstack/react-query'
```

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <File.Import name={['useQueryClient']} path="@tanstack/react-query" />
+        <File.Import name={['HookOptions']} root={file.path} path={path.resolve(root, './index.ts')} />
+        <File.Source name={file.name}>
```
stijnvanhulle: We need some extra prop to make this work with the barrel generation

```
 <File.Source name={file,name} isExportable isIndexable>
```

stijnvanhulle:
@JoaoBrlt Seems reasonable, let's already add `operationId` and `hookName` so both cases will be covered.

stijnvanhulle:
@all-contributors please add @JoaoBrlt for code

packages/core/src/types.ts
JoaoBrlt: If I understood correctly, the `output` option of each plugin uses this `Output` type.
Indeed, the `output` option of the plugins seems to be always be typed as `Output<Oas>`.
So we can already set the `output.override` per plugin. :smile: 

What I didn't manage to do is set the `output.override` option of all plugins when setting the `output.override` option at the root of the `kubb.config.ts` file.
But maybe that's not even required.

packages/plugin-react-query/src/generators/hookOptionsGenerator.tsx
JoaoBrlt: Good point :ok_hand: I've just pushed a commit to do that.

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
JoaoBrlt: Good catch :ok_hand: I didn't know this option. It should be fixed now.

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
JoaoBrlt: Good point 👌 I've just pushed a commit to do that.

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const { importPath, name } = options.customOptions
+
+    const root = path.resolve(config.root, config.output.path)
+
+    const hookIndexFile = pluginManager.getFile({ name: 'index', extname: '.ts', pluginKey })
+
+    const file = {
+      baseName: path.basename(importPath) as `${string}.${string}`,
+      name: path.basename(importPath, path.extname(importPath)),
+      path: path.resolve(hookIndexFile.path, importPath),
```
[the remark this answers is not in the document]
JoaoBrlt: We don't manipulate the path when using it in the generated React Query hook files.
However, when generating the default custom hook options file, we need to use the hook files as the base path to generate the file in the appropriate location. :wink:

packages/core/src/types.ts
stijnvanhulle: @JoaoBrlt I would say to use `output.override` of the plugin first and fallback on `ouput.override` of the config (main/root) one. I think we did the same thing for `barrelType`

JoaoBrlt: Ah, I just understood. The global config is the one passed to the generator function directly (in contrast to the one retrieved from the `plugin` variable). I'll do that then. :ok_hand:

JoaoBrlt: It should be done now. :wink:

packages/core/src/types.ts
```diff
@@ -328,6 +328,11 @@ export type Output<TOptions> = {
...
+   * @default false
+   */
+  override?: boolean
```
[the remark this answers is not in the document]
stijnvanhulle: @JoaoBrlt This one can be removed, no?

stijnvanhulle:
@JoaoBrlt Before we could finish this PR, we need an updated example (maybe one to invalidate some queries, a feature that was requested a couple of times), docs and updated changelog(`pnpm changeset`), any thing else we still need before merging this feature? 

Happy new year btw 💯

JoaoBrlt:
Hey @stijnvanhulle Happy new year! :tada: Great, I was waiting for your feedback. I'll do that. :ok_hand: Do you prefer to update the existing `react-query` example or to add a new `react-query-custom-options` example? Otherwise, I don't see anything else.

stijnvanhulle:
@JoaoBrlt I would stay to update the existing react-query example, I have been working on a new tests folder with more test cases so when finished it could also be moved to there.

JoaoBrlt:
Hi @stijnvanhulle! :wave: So I've updated the `react-query` example, added documentation for the `customOptions` option (react-query plugin only) and `output.override` option (core and all plugins) and added a changeset. :wink: Don't hesitate to tell me if something is wrong or missing.

packages/plugin-react-query/src/components/Query.tsx
```diff
@@ -164,9 +166,10 @@ export function Query({
        const { query: queryConfig = {}, client: config = {} } = options ?? {}
        const { client: queryClient, ...queryOptions } = queryConfig
        const queryKey = queryOptions?.queryKey ?? ${queryKeyName}(${queryKeyParams.toCall()})
+       ${customOptions ? `const customOptions = ${customOptions.name}({ hookName: '${name}', operationId: '${operation.getOperationId()}' })` : ''}
 
        const query = useQuery({
-        ...${queryOptions},
+        ...${queryOptions},${customOptions ? '\n...customOptions,' : ''}
         queryKey,
         ...queryOptions
```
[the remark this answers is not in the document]
JoaoBrlt: I think the order is correct. 
Indeed, I think :
- Custom options should override base options (in order to be able to customize most properties),
- Custom options should NOT override the query key (the query key can already be customized using another Kubb option which also handle the query key generator function),
- Query options should override custom options (in order to be able to override custom options if needed in some specific places).

packages/plugin-react-query/src/components/Mutation.tsx
```diff
@@ -140,9 +142,10 @@ export function Mutation({
         const mutationKey = mutationOptions.mutationKey ?? ${mutationKeyName}(${mutationKeyParams.toCall()})
 
         const baseOptions = ${mutationOptions} as UseMutationOptions<${generics}>
+        ${customOptions ? `const customOptions = ${customOptions.name}({ hookName: '${name}', operationId: '${operation.getOperationId()}' }) as UseMutationOptions<${generics}>` : ''}
 
         return useMutation<${generics}>({
-          ...baseOptions,
+          ...baseOptions,${customOptions ? '\n...customOptions,' : ''}
           mutationKey,
           ...mutationOptions,
```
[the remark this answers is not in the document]
JoaoBrlt: I think the order is correct. 
Indeed, I think :
- Custom options should override base options (in order to be able to customize most properties),
- Custom options should NOT override the query key (the query key can already be customized using another Kubb option which also handle the query key generator function),
- Mutation options should override custom options (in order to be able to override custom options if needed in some specific places).

packages/plugin-react-query/src/components/SuspenseQuery.tsx
```diff
@@ -171,9 +173,10 @@ export function SuspenseQuery({
        const { query: queryConfig = {}, client: config = {} } = options ?? {}
        const { client: queryClient, ...queryOptions } = queryConfig
        const queryKey = queryOptions?.queryKey ?? ${queryKeyName}(${queryKeyParams.toCall()})
+       ${customOptions ? `const customOptions = ${customOptions.name}({ hookName: '${name}', operationId: '${operation.getOperationId()}' })` : ''}
 
        const query = useSuspenseQuery({
-        ...${queryOptions},
+        ...${queryOptions},${customOptions ? '\n...customOptions,' : ''}
         queryKey,
         ...queryOptions
```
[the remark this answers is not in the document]
JoaoBrlt: I think the order is correct. 
Indeed, I think :
- Custom options should override base options (in order to be able to customize most properties),
- Custom options should NOT override the query key (the query key can already be customized using another Kubb option which also handle the query key generator function),
- Query options should override custom options (in order to be able to override custom options if needed in some specific places).

packages/plugin-react-query/src/components/InfiniteQuery.tsx
```diff
@@ -193,9 +195,10 @@ export function InfiniteQuery({
        const { query: queryConfig = {}, client: config = {} } = options ?? {}
        const { client: queryClient, ...queryOptions } = queryConfig
        const queryKey = queryOptions?.queryKey ?? ${queryKeyName}(${queryKeyParams.toCall()})
+       ${customOptions ? `const customOptions = ${customOptions.name}({ hookName: '${name}', operationId: '${operation.getOperationId()}' })` : ''}
 
        const query = useInfiniteQuery({
-        ...${queryOptions},
+        ...${queryOptions},${customOptions ? '\n...customOptions,' : ''}
         queryKey,
         ...queryOptions
```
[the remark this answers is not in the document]
JoaoBrlt: I think the order is correct. 
Indeed, I think :
- Custom options should override base options (in order to be able to customize most properties),
- Custom options should NOT override the query key (the query key can already be customized using another Kubb option which also handle the query key generator function),
- Query options should override custom options (in order to be able to override custom options if needed in some specific places).

packages/plugin-react-query/src/components/SuspenseInfiniteQuery.tsx
```diff
@@ -193,9 +195,10 @@ export function SuspenseInfiniteQuery({
        const { query: queryConfig = {}, client: config = {} } = options ?? {}
        const { client: queryClient, ...queryOptions } = queryConfig
        const queryKey = queryOptions?.queryKey ?? ${queryKeyName}(${queryKeyParams.toCall()})
+       ${customOptions ? `const customOptions = ${customOptions.name}({ hookName: '${name}', operationId: '${operation.getOperationId()}' })` : ''}
 
        const query = useSuspenseInfiniteQuery({
-        ...${queryOptions},
+        ...${queryOptions},${customOptions ? '\n...customOptions,' : ''}
         queryKey,
         ...queryOptions
```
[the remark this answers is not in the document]
JoaoBrlt: I think the order is correct. 
Indeed, I think :
- Custom options should override base options (in order to be able to customize most properties),
- Custom options should NOT override the query key (the query key can already be customized using another Kubb option which also handle the query key generator function),
- Query options should override custom options (in order to be able to override custom options if needed in some specific places).

packages/plugin-react-query/src/generators/customHookOptionsFileGenerator.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <Function
+            name={name}
+            generics="T extends keyof HookOptions"
+            params="{ hookName, operationId }: { hookName: T, operationId: string }"
+            returnType="HookOptions[T]"
+            export
+          >
+            {`const queryClient = useQueryClient()
+            const customOptions = getCustomHookOptions({ queryClient })
+            return customOptions[hookName] ?? {}`}
```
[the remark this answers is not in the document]
JoaoBrlt: The `operationId` parameter is not used in the generated example because the generated example uses the type-safe approach (using the generated `HookOptions` type).
However, the `operationId` is always provided in case developers want to customize the hook options depending on the operation id coming from the OpenAPI specs.

stijnvanhulle:
@JoaoBrlt Thanks a lot for creating this feature, I will merge and create new version. Looking forward in what the community will think about it.

PR #2209, opened 2025-12-18
Add params to Cypress requests

Ericlm:
## 🎯 Changes

This PR adds a way for the Cypress plugin to set params type (path params, query params…).
It is very close to the plugin client in this way.

## ✅ Checklist

- [X] I have tested this code locally with `pnpm run test`.

## 🚀 Release Impact


packages/plugin-cypress/src/types.ts
```diff
@@ -22,6 +22,25 @@ export type Options = {
    * @default 'data'
    */
   dataReturnType?: 'data' | 'full'
+  /**
```
stijnvanhulle, line 25: @Ericlm Could you also have a look at adding this to the cypress docs?

PR #2316, opened 2026-01-15
feat(cli): add oxfmt formatter support

EricPierlotIdmog:
## 🎯 Changes

Add [oxfmt](https://github.com/user/oxfmt) as a new formatter option for generated files.

### What's included:
- Add oxfmt to formatters config (`packages/cli/src/utils/formatters.ts`)
- Update `detectFormatter` to detect oxfmt with priority: biome > oxfmt > prettier
- Update types to include `'oxfmt'` option in `format` config
- Update documentation

### Usage:
```ts
// kubb.config.ts
export default defineConfig({
  output: {
    format: 'oxfmt', // or 'auto' to auto-detect
  },
})
```

### Why oxfmt?
Oxfmt is a fast Rust-based formatter (part of the oxc toolchain, like oxlint which is already supported). Adding it provides consistency for users already using oxlint for linting.

## ✅ Checklist

## 🚀 Release Impact

## 💡 Follow-up
I noticed the linter section uses the same pattern (repeated if-blocks for eslint/biome/oxlint). Happy to submit a follow-up PR to refactor both formatters and linters into a unified declarative config if you're interested !

### DIFF
--- a/.changeset/cool-monkeys-enter.md
+++ b/.changeset/cool-monkeys-enter.md
@@ -1,0 +1,6 @@
+---
+"@kubb/core": minor
+"@kubb/cli": minor
+---
+
+Add oxfmt as a new formatter option
--- a/packages/cli/src/utils/formatters.ts
+++ b/packages/cli/src/utils/formatters.ts
@@ -1,0 +1,17 @@
+export const formatters = {
+  prettier: {
+    command: 'prettier',
+    args: (outputPath: string) => ['--ignore-unknown', '--write', outputPath],
+    errorMessage: 'Prettier not found',
+  },
+  biome: {
+    command: 'biome',
+    args: (outputPath: string) => ['format', '--write', outputPath],
+    errorMessage: 'Biome not found',
+  },
+  oxfmt: {
+    command: 'oxfmt',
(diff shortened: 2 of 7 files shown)


packages/cli/src/utils/formatters.ts
```diff
@@ -0,0 +1,17 @@
+export const formatters = {
+  prettier: {
```
stijnvanhulle, line 2: Not part of this feature but we could do the same for the linters to reduce code

stijnvanhulle (approved):

.changeset/cool-monkeys-enter.md
```diff
@@ -0,0 +1,6 @@
+---
+"@kubb/core": major
```
stijnvanhulle: Please use minor instead, no need to have a v5 of Kubb for this

stijnvanhulle (changes requested):

packages/cli/src/utils/formatters.ts
EricPierlotIdmog: Thanks for the review ! 
I'd be happy to submit a follow-up PR to refactor the linters the same way. I'll open it once this one is merged.

PR #2333, opened 2026-01-20
Feature: Add support for static class client generation

### LINKED ISSUE #2326: Feature Request: Static methods in class client
> [!NOTE]
> Originally posted as discussion #2322, moved here.

Some other libraries like Kubb allow the generation of class-clients/services with static members, allowing us to change the usage:
```diff
- new TestingService().doThing()
+ TestingService.doThing()
```

This can be achieved by simply adding an option for `pluginClient().clientType='staticClass'` (or something), which would change generated client signature as follows:

```diff
export class TestService {
-  #client: typeof fetch
+  static #client: typeof fetch

-   async doTestAction(testId: TestActionParam["testId"], data: TestActionMutationRequest, config: Partial<RequestConfig<TestActionMutationRequest>> & { client?: typeof fetch } = {}) {
+   static async doTestAction(testId: TestActionParam["testId"], data: TestActionMutationRequest, config: Partial<RequestConfig<TestActionMutationRequest>> & { client?: typeof fetch } = {}) {
    ...
```

And other necessary changes, like how references to `this` must be replaced by `TestService`. For example, replace `this.#client` with `StepService.#client` (and potentially make it non-private, not sure how best-practice is for this).

sebastianvitterso:
## 🎯 Changes

Added a new `clientType: 'staticClass'` option to generate API clients as classes with static methods. Updated documentation to reflect this feature and included examples for usage.

Closes #2326.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/docs/knowledge-base/debugging.md
+++ b/docs/knowledge-base/debugging.md
@@ -39,9 +39,9 @@ npm install --save-dev @kubb/react-fabric
 ```shell [yarn]
 yarn add -D @kubb/react-fabric
 ```
+:::
 
 ### Update `kubb.config.ts`
-:::
 
 ```typescript{1}
 import { openDevtools } from '@kubb/react-fabric' // [!code ++]


sebastianvitterso:
Please note that I've "vibe coded" 90% of this, and as verification, all I know is that it looks a lot like the regular `class` client, with the seemingly correct exceptions. 

Let me know if anything is off, or feel free to fix small stuff yourself. I've opened up the PR for editing by maintainers @stijnvanhulle.

docs/plugins/plugin-client.md
```diff
@@ -157,8 +157,118 @@ export function getGetPetByIdUrl(petId: GetPetByIdPathParams['petId']) {
 ### client
 <!--@include: ./plugin-client/client.md-->
 
-### clientType <span class="new">new in 4.9.0</span>
-<!--@include: ./plugin-client/clientType.md-->
+### clientType <span class="new">new in 4.18.0</span>
+
+How to generate the client code
```
stijnvanhulle: Can't we just use `<!--@include: ./plugin-client/clientType.md-->`, this will import the clientType.md file and render it inside of this header

packages/plugin-client/src/components/StaticClassClient.tsx
```diff
@@ -0,0 +1,224 @@
+import { buildJSDoc, URLPath } from '@kubb/core/utils'
```
stijnvanhulle, line 1: Can you also expose this in `packages/plugin-client/src/components/index.ts`

sebastianvitterso:
Addressed your comments! 😄 

As I said, there was a lot of copilot in use when I made this, as well as basing it on the existing `ClassClient` one, but the fabric-system looks very interesting! I've always tried to steer away from code generation systems, as they always seem so advanced, but I think using React as a basis is a cool approach.

Considering that kubb has come as far as it has, it wasn't the easiest case to come aboard fabric with, considering "how much" the generator already had to implement. But copilot and copying helped a lot, which merits kudos for making a system structured enough with sufficient documentation to be understandable for an LLM.

stijnvanhulle:
@sebastianvitterso My guess is that the LLM is smart enough is probably because it understands JSX syntax (functional components) so easier for the LLM to view some examples. Also thinking about adding agent skills for Kubb so LLMs could even understand more (kind of docs for llms but local). 

Fabric also needs some improvements so quite busy the last couple of days to improve the render engine. What is also quite interesting is that we could use the `react-devtools` to debug the components (viewing only): https://kubb.dev/kubb/knowledge-base/debugging#react-devtools

PR #2367, opened 2026-01-23
feat: orpc contract plugin

hyoban:
## 🎯 Changes

oRPC features a Contract-first mode that allows contracts to be defined in advance that comply with the OpenAPI specification, followed by separate implementations for the frontend and backend.

In this PR, we have introduced the oRPC plugin to generate oRPC Contracts from OpenAPI. This is useful for projects that do not use TypeScript with an oRPC backend.

- https://orpc.dev/docs/contract-first/define-contract
- https://orpc.dev/docs/openapi/getting-started

If you don't want to introduce this plugin, please feel free to close this PR.

To pass the CI, we will need #2368.

## ✅ Checklist

## 🚀 Release Impact


nasoooor29:
hello @hyoban thanks for the great plugin i really liked it 

> i just knew about kubb today so i spent the whole day tinkering it was perfect

i just have a request (if it's fine with you), can you make the extention make something like a subrouter based on the tags?
maybe something like this
> to be honest it's just a preference thing but if you can do it it would be great 🫡

https://github.com/nasoooor29/orpc-contract-creator/blob/main/generated/visory/contract.ts#L1096-L1172

# sorry i didn't know it was already implemented i didn't see it on the example of the changed files inside the PR

packages/plugin-orpc/package.json
```diff
@@ -0,0 +1,97 @@
...
+    "typecheck": "tsc -p ./tsconfig.json --noEmit --emitDeclarationOnly false"
+  },
+  "dependencies": {
+    "@kubb/core": "workspace:*",
+    "@kubb/oas": "workspace:*",
+    "@kubb/plugin-oas": "workspace:*",
+    "@kubb/plugin-zod": "workspace:*",
+    "@kubb/react-fabric": "catalog:"
+  },
+  "devDependencies": {
```
stijnvanhulle, line 86: We need peer dependency for @kubb/react-fabric here lik we have for other plugins

packages/plugin-orpc/src/types.ts
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Generate a router file that groups all contracts by tag.
+   * @default false
+   */
+  router?: boolean
+  /**
+   * Name of the exported router object.
+   * @default 'router'
+   */
+  routerName?: string
```
stijnvanhulle, line 59: Will this mean only one router is possible per openapi file? That is not per router/definitionId?

stijnvanhulle (changes requested):

stijnvanhulle:
We will also need some docs so new users can understand what the orpc plugin is about. Also have not yet used Orpc so not 100% sure if everything will work fine. I will try the examples to see what it is about. Thanks for the request and the implementation of this new plugin, approach this a lot!

hyoban:
Thank you very much for your review. There are still some shortcomings in this PR, and I am happy to fix them and add documentation.

stijnvanhulle:
@hyoban thanks, let me know if you need any help here. Also we did some changes so best to pull main and do another pnpm i to continue.

PR #2427, opened 2026-02-05
fix: externalize @kubb/core in tsdown configs

icholy:
tsdown inlines type declarations from dependencies that aren't marked as external. This causes each package to bundle its own copy of PluginManager (which has #private fields), making TypeScript treat them as incompatible types across packages.

Adding '@kubb/core' to the external list makes tsdown emit import references instead of inlining, so all packages share the same PluginManager.

I ran into this while trying to write a custom generator plugin.

## 🎯 Changes

Add `@kubb/core` to externals in all tsdown configurations.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/plugin-ts/src/components/Type.tsx
+++ b/packages/plugin-ts/src/components/Type.tsx
@@ -166,7 +166,7 @@ export function Type({
         enums.map(({ name, nameNode, typeName, typeNode }) => (
           <>
             {nameNode && (
-              <File.Source name={name} isExportable isIndexable>
+              <File.Source name={name} isExportable isIndexable isTypeOnly={false}>
                 {safePrint(nameNode)}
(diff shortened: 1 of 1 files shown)


stijnvanhulle:
@icholy hmm I thought that adding `@kubb/core` to the dependencies should be enough and that `external` is an override?

```
 "dependencies": {
    "@kubb/core": "workspace:*",
    "@kubb/oas": "workspace:*",
    "@kubb/plugin-oas": "workspace:*",
    "@kubb/plugin-ts": "workspace:*",
    "@kubb/plugin-zod": "workspace:*",
    "@kubb/react-fabric": "catalog:"
  },
```

stijnvanhulle:
@all-contributors please add @icholy for code

PR #2463, opened 2026-02-10
fix: add mergeConfig to fetch/axios clients and use it in class clients

icholy:
## 🎯 Changes

Constructor configs passed to class-style API clients were silently discarded. Only the `client` property was stored, all other `RequestConfig` properties like `baseURL` and `headers` were ignored.

- Add `mergeConfig` export to both fetch and axios client templates
- Use `mergeConfig` in generated class client methods to merge stored constructor/class config with per-method config
- Use `mergeConfig` internally in the fetch and axios client functions to replace inline merge logic

I originally tried to just code-gen the merging code, but since fetch allows the headers to an object or array that was really gross. Another way would be to codegen a `#mergeConfig` method.

## ✅ Checklist

## 🚀 Release Impact


icholy:
@stijnvanhulle thanks for improving that logging. I'm putting this back into draft while I figure this out.

icholy:
@stijnvanhulle ~~I'm running into issues because of the drift between `src/clients/{fetch,axios}.ts` and `templates/clients/{fetch,axios}.ts`. I think I want to open a separate PR to sync these before contining on this one.~~

 **edit:** I just changed the import style instead.

icholy:
@stijnvanhulle this is ready for a review when you have time.

PR #2464, opened 2026-02-10
fix: only import Client type when inline client is rendered

icholy:
## 🎯 Changes

Only include `Client` in the type-only import when `!shouldUseClientPlugin`, since `Client` is only referenced by the inline async client function.

## ✅ Checklist

## 🚀 Release Impact


icholy:
This is not correct.

PR #2394, opened 2026-01-29
feat(faker): data forwarding

Ericlm:
## 🎯 Changes

Closes #2393 

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/kubb-labs/kubb/blob/main/CONTRIBUTING.md).
- [x] I have tested this code locally with `pnpm run test`.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [x] This change is for the docs (no release).

PR #2472, opened 2026-02-12
fix: forwarding function from enum string

Ericlm:
## 🎯 Changes

A previous change with how faker handled functions was making it possible to pass data downstream. However, in some edge case where the function that was called was an enum, no data was accepted (it wouldn't make sense). But faker functions were still calling this function with data. (see https://github.com/kubb-labs/kubb/pull/2394)

Now, functions generated from enums are also accepting data.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/core/mocks/helloWorld.js
+++ b/packages/core/mocks/helloWorld.js
@@ -1,1 +1,0 @@
-export const hallo = 'world'


packages/core/mocks/helloWorld.js
```diff
@@ -0,0 +1 @@
+export const hallo = 'world'
```
stijnvanhulle: Normally this should be ignored by the gitignore file

stijnvanhulle:
@Ericlm could you also check the changelog.md file in docs and add already what new version will be 4.22.2

packages/core/mocks/helloWorld.js
Ericlm: Should I add `packages/core/mocks/hellowWorld.js` to the gitignore? Or remove one of `packages/core/mocks/hellowWorld.js` VS `packages/core/mocks/helloWorld.js`

stijnvanhulle: It is already part of the gitignore so not sure if this because merge conflicts or it stills happens

See https://github.com/kubb-labs/kubb/blob/main/.gitignore.

Both files should not be part of git

stijnvanhulle (approved):

packages/core/mocks/helloWorld.js
Ericlm: My bad! I was able to remove this file from the git tracking. 🙂

PR #2475, opened 2026-02-13
fix(plugin-react-query): respect required requestBody for oneOf mutation inputs

humarkx:
## Summary
- fix required request body inference in `plugin-react-query` mutations when the request schema is a top-level `oneOf`
- require mutation `data` when the OpenAPI operation has `requestBody.required: true`, even if schema-level optionality is ambiguous
- preserve compatibility with `plugin-client` by keeping client call parameter ordering aligned

## Problem
For some operations, `requestBody.required` is `true` but the request schema is a top-level `oneOf` without top-level `required`.

That caused generated mutation typings to incorrectly allow `data?`.
A first pass also exposed a mismatch when `plugin-client` is enabled: mutation wrappers and imported client functions could disagree on parameter ordering (e.g. header params vs request body).

## Fix
- in `Mutation` and `MutationOptions`, derive request-body optionality from operation-level `requestBody.required`
- in `mutationGenerator`, only normalize request schema optionality for inline clients (when `plugin-client` is not used)
- when `plugin-client` is used, keep original schema shape for client-call argument mapping so generated calls stay type-safe

## Validation
- `pnpm --filter @kubb/plugin-react-query test`
- `pnpm --filter @kubb/plugin-react-query typecheck`
- PR CI: both E2E jobs (`ubuntu-latest`, `windows-latest`) pass


packages/plugin-react-query/src/components/MutationOptions.tsx
```diff
@@ -22,6 +22,12 @@ type GetParamsProps = {
   typeSchemas: OperationSchemas
 }
 
+function isRequestBodyRequired(typeSchemas: OperationSchemas) {
```
stijnvanhulle: @humarkx could you move this to the utils ofr @kubb/plugin-oas package?

packages/plugin-react-query/src/generators/mutationGenerator.tsx
```diff
@@ -13,6 +14,29 @@ import { Mutation, MutationKey } from '../components'
 import { MutationOptions } from '../components/MutationOptions.tsx'
 import type { PluginReactQuery } from '../types'
 
+function withRequiredRequestBodySchema(typeSchemas: OperationSchemas): OperationSchemas {
```
stijnvanhulle: Would it make sense to add this as part of the schema generation so every plugin can read this in correctly?

humarkx:
@stijnvanhulle can this PR get some love?

stijnvanhulle (changes requested):

packages/plugin-react-query/src/generators/mutationGenerator.tsx
hmrkx: Applied with the same refactor: required-requestBody normalization now happens in schema generation (`OperationGenerator`) via shared `@kubb/plugin-oas` utility, so plugins read consistent requiredness automatically. I also added regression coverage for react-query (inline + plugin-client path) and plugin-client generation with required oneOf requestBody.

packages/plugin-react-query/src/components/MutationOptions.tsx
humarkx: Implemented in this update. I moved required requestBody handling to shared `@kubb/plugin-oas` utils (`requestBody.ts`) and wired it through `OperationGenerator.getSchemas()`. `plugin-react-query` now consumes the normalized schema and no longer carries a local helper.

packages/plugin-react-query/src/generators/mutationGenerator.tsx
humarkx: Applied with the same refactor: required requestBody normalization now happens in schema generation (`OperationGenerator`) via shared `@kubb/plugin-oas` utility, so plugins read consistent requiredness automatically. Added regression coverage for react-query (inline + plugin-client path) and plugin-client generation with required oneOf requestBody.

humarkx:
@stijnvanhulle we’re running this in production via `patch-package` and it’s been solid so far.

It avoids having to deal with `data?` when the response body is actually required.
Also handles `oneOf` / `anyOf` / `allOf` composable required-body cases correctly.

stijnvanhulle:
@humarkx Thanks for fixing merge feedback, I will create a new version of Kubb with this fix :)

PR #2484, opened 2026-02-17
Agent Websocket integration

stijnvanhulle:
## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/agent-websocket-integration.md
+++ b/.changeset/agent-websocket-integration.md
@@ -1,0 +1,19 @@
+---
+'@kubb/agent': minor
+'@kubb/cli': patch
+---
+
+WebSocket integration for Kubb Studio connectivity
+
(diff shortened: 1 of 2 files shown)


packages/agent/server/utils/sessionManager.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+/**
+ * Hash agent token for use as session key
+ * This prevents storing raw tokens in the config file
+ */
+function hashAgentToken(token: string): string {
+  // Use a simple hash - in production you might want to use crypto
+  return Buffer.from(token).toString('base64').substring(0, 16)
```
stijnvanhulle: @copilot  use of 

```
import { createHash, randomBytes } from 'node:crypto'

/**
 * Hash an agent token using SHA-512
 * Tokens are hashed before storage for security - raw tokens are never persisted
 */
export function hashAgentToken(token: string): string {
  return createHash('sha512').update(token).digest('hex')
}
```
instead

stijnvanhulle:
@copilot add changelog and changeset

PR #2588, opened 2026-02-25
fix: don't mutate config.signal instead use it only if available

jay-babu:
## 🎯 Changes

When using the retry config and NOT specifying a `signal`, a new signal is constructed. When the signal is mutated, it is only done once. Previously this wasn't a problem since config.signal was mutated every time. This change simply says don't mutate config every, just set signal if not given.

## ✅ Checklist

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).


jay-babu:
@stijnvanhulle i am trying regenerate the code but it is not working. unsure what I am doing incorrectly.

stijnvanhulle:
@jay-babu what kind of errors/issues to you get?

jay-babu:
> @jay-babu what kind of errors/issues to you get?

Not getting any errors when generating but the CI test is failing from looks like some missing generated code?

stijnvanhulle:
@jay-babu I see that your test cases are not up to date, so probably just run `pnpm test u` see https://vitest.dev/guide/snapshot.html

stijnvanhulle (approved):

PR #2751, opened 2026-03-15
fix(plugin-ts): prevent AST crash on negative literal enum values

### LINKED ISSUE #2734: @kubb/plugin-ts literal enums crash on negative numeric values

4.33.3

MacOS

```typescript
import { defineConfig } from '@kubb/core';

export default defineConfig({
  input: {
    path: './openapi.json',
  },
  output: {
    path: './dist',
  },
  plugins: [
    pluginOas(),
    pluginTs({
      output: {
        path: './types',
      },
      enumType: 'literal',
    }),
  ],
});
```

### Swagger/OpenAPI file?

```bash
{
  "openapi": "3.0.0",
  "info": {
    "title": "repro",
    "version": "1.0.0"
  },
  "paths": {},
  "components": {
    "schemas": {
      "Allowance": {
        "type": "number",
        "enum": [-1, 0, 5]
      }
    }
  }
}
```

```JSON

```

Minimal reproduction at https://stackblitz.com/edit/vitejs-vite-tywmdngq?file=kubb.config.ts&view=editor

Every time

Valid generated TypeScript for the enum, for example `type Allowance = -1 | 0 | 5`

### Additional information

-  switching to enumType: 'asConst' avoids the crash
-  this suggests the bug is specific to the literal enum AST generation path
-  the likely fix is to emit negative numeric literals with createPrefixUnaryExpression instead of createNumericLiteral(-1)

Sigmabrogz:
## Summary
Fixes an issue where `@kubb/plugin-ts` crashes when generating literal enums that contain negative numeric values.

## Problem
Fixes #2734.
When `enumType: 'literal'` or `'inlineLiteral'` is used and an enum contains a negative number (e.g. `enum: [-1, 0, 5]`),  invokes `factory.createNumericLiteral(value?.toString())`. Passing a negative string (like `"-1"`) to `createNumericLiteral` is an invalid AST construct in the TypeScript compiler API, causing the build to crash or output malformed code.

## Solution
Updated `factory.ts` to wrap negative numeric values inside a `PrefixUnaryExpression` using `ts.SyntaxKind.MinusToken` and the absolute value, ensuring a valid and compilable TypeScript AST.

### DIFF
--- a/packages/plugin-ts/src/factory.ts
+++ b/packages/plugin-ts/src/factory.ts
@@ -481,7 +481,13 @@ export function createEnumDeclaration({
           enums
             .map(([_key, value]) => {
               if (isNumber(value)) {
-                return factory.createLiteralTypeNode(factory.createNumericLiteral(value?.toString()))
+                if (value < 0) {
+                  return factory.createLiteralTypeNode(
+                    factory.createPrefixUnaryExpression(ts.SyntaxKind.MinusToken, factory.createNumericLiteral(Math.abs(value).toString()))
+                  )
+                } else {
+                  return factory.createLiteralTypeNode(factory.createNumericLiteral(value?.toString()))
+                }
               }
 
               if (typeof value === 'boolean') {


Sigmabrogz:
This fixes the AST crash by wrapping negative literal numeric enum values in a `PrefixUnaryExpression`, addressing #2734.

PR #2759, opened 2026-03-16
Custom storage as memoryStorage and fsStorage with possibility to save to s3 buckets

stijnvanhulle:
## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/storage-abstraction.md
+++ b/.changeset/storage-abstraction.md
@@ -9,7 +9,7 @@ Introduces a `storage` option in `output` that replaces direct filesystem writes
 **New exports from `@kubb/core`:**
 
 - `defineStorage(builder)` — factory helper (same pattern as `definePlugin`/`defineLogger`/`defineAdapter`) that wraps a builder function and makes options optional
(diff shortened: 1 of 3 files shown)


mbrevda:
anywhere to see the backstory on this pr? I'm wondering why adding a bunch of storage adapters to a code generator is a good idea (i.e., would be better to leave that for other, dedicated, tools, no?)? Or maybe I'm misunderstanding the intent of this? Either way, if you can point to an issue/discussion, that would be great

stijnvanhulle:
@mbrevda This is for now only for internal usage, this could be used to safe files directly to s3 or anything else but today this is added as part of our new agent package (more to come soon). 

The main idea is here to make it possible to safe the files to whatever storage you like (idea from https://unstorage.unjs.io/). so Kubb does not really need to know where (up to the user to define this). To clarify, we ware not planning to add s3 or another storage adapters available but one generic way of defining it will be part of Kubb. Does this answer your question?

mbrevda:
Yes, thanks! ❤️ 
(for the future: perhaps adding just another line or two to the change log could help keep things clearer)

PR #2881, opened 2026-03-23
fix: Remove the `collisionDetection` option from `adapterOas` and always use collision-safe schema/enum naming in v5.

stijnvanhulle:
## ✅ Checklist

## 🚀 Release Impact


packages/plugin-ts/src/presets.ts
```diff
@@ -0,0 +1,57 @@
+import type { Visitor } from '@kubb/ast/types'
+import { mergeResolvers } from '@kubb/core'
+import { resolverTs, resolverTsHeyapi, resolverTsLegacy, resolverTsOrval } from './resolvers/index.ts'
+import type { CompatibilityPreset, ResolverTs } from './types.ts'
+
+export type TsCompatibilityPreset = {
```
stijnvanhulle: @copilot add preset to the core and create a new definePreset (name, resolvers, transformers) to make it as easy as possible to create new presets

packages/plugin-ts/src/types.ts
```diff
@@ -4,6 +4,8 @@ import type { contentType, Oas } from '@kubb/oas'
 import type { Exclude, Include, Override, ResolvePathOptions } from '@kubb/plugin-oas'
 import type { Generator } from '@kubb/plugin-oas/generators'
 
+export type CompatibilityPreset = 'none' | 'kubbV4' | 'heyapi' | 'orval'
```
stijnvanhulle: @copilot move this type to the core as we will reuse this in other plugins

PR #3175, opened 2026-04-25
fix: preserve oneOf/anyOf distinction in union schema nodes

stijnvanhulle:
Add unionKind field to UnionSchemaNode to track whether a union comes from
oneOf or anyOf. This allows plugins to correctly validate oneOf constraints
(exactly one member must match) vs anyOf semantics (any number can match).

- Add 'unionKind' field to UnionSchemaNode AST type
- Update parser to set unionKind based on which field is present
- Add tests to verify unionKind is set correctly
- Prioritize oneOf when both are present

https://claude.ai/code/session_01A1PTcR3skoeRd9oLsJJqjh

### DIFF
--- a/docs/union-plan.md
+++ b/docs/union-plan.md
@@ -45,13 +45,13 @@ const invalidData = {
 
 ### 1. AST Enhancement
 
-Added `mode` field to `UnionSchemaNode` in `packages/ast/src/nodes/schema.ts`:
+Added `operator` field to `UnionSchemaNode` in `packages/ast/src/nodes/schema.ts`:
 
 ```typescript
 export type UnionSchemaNode = CompositeSchemaNodeBase & {
   type: 'union'
   discriminatorPropertyName?: string
-  mode?: 'one' | 'any'  // NEW: 'one' = oneOf (exactly one), 'any' = anyOf (any number)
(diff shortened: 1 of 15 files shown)


docs/union-plan.md
```diff
@@ -0,0 +1,149 @@
...
+## Solution
+
+### 1. AST Enhancement
+
+Added `unionKind` field to `UnionSchemaNode` in `packages/ast/src/nodes/schema.ts`:
+
+```typescript
+export type UnionSchemaNode = CompositeSchemaNodeBase & {
+  type: 'union'
+  discriminatorPropertyName?: string
```
stijnvanhulle, line 53: @copilot can we use another approach so we are not exposing oneOf, anyOf. Find better generic name

stijnvanhulle: @copilot find better name for unionKind

stijnvanhulle: @copilot use simplere names in english

stijnvanhulle: @copilot better ast related name

stijnvanhulle: @copilot use strategy as name

stijnvanhulle: @copilot update the 2 plan md files to reflect this name change

PR #3176, opened 2026-04-25
fix(middleware-barrel): prevent banner/footer from bleeding into barrel files

### LINKED ISSUE #2265: Feature Request: Option to exclude barrel files from `output.banner`
# Feature Request: Option to exclude barrel files from `output.banner`

## Problem

When using `pluginClient` with `output.banner: "'use server'"` for Next.js Server Actions, **all generated files** receive this directive, including:

1. **Barrel files** (`index.ts`) - only do re-exports
2. **Group aggregation files** (`stocks/stocks.ts`, `broker/broker.ts`, etc.) - generated when using `group` option

These files should NOT have `"use server"` because:
- Barrel files only re-export, they don't define server actions
- Aggregation files return function references which cannot be serialized in server actions

## Current Workaround

We have to use `hooks.done` with a post-processing script to remove `"use server"` from barrel files:

```typescript
// kubb.config.ts
export default defineConfig({
  hooks: {
    done: ['bash scripts/remove-use-server-from-barrels.sh'],
  },
  plugins: [
    pluginClient({
      output: {
        path: './api',
        banner: "'use server'",
        barrelType: 'named',
      },
      group: { type: 'tag', name: ({ group }) => camelCase(group) },
    }),
  ],
});
```

```bash
# remove-use-server-from-barrels.sh
# Remove "use server" from dir/dir.ts pattern files
find src/actions-generated/api -regex '.*/\([^/]*\)/\1\.ts' -exec perl -i -pe 's/"use server";//' {} \;
```

## Proposed Solution

Add an option to exclude barrel/aggregation files from banner injection:

### Option A: `bannerExcludeBarrel`

```typescript
pluginClient({
  output: {
    path: './api',
    banner: "'use server'",
    bannerExcludeBarrel: true,  // Don't add banner to index.ts and dir/dir.ts files
  },
}),
```

### Option B: `banner` as function with file context

```typescript
pluginClient({
  output: {
    path: './api',
    banner: ({ isBarrel, isAggregation, filePath }) => {
      if (isBarrel || isAggregation) return '';
      return "'use server'";
    },
  },
}),
```

### Option C: Separate `barrelBanner` option

```typescript
pluginClient({
  output: {
    path: './api',
    banner: "'use server'",
    barrelBanner: '',  // Override banner for barrel files
  },
}),
```

## Environment

- Kubb version: 4.12.15
- Framework: Next.js 16.1 (App Router with Server Actions)
- Plugins: `@kubb/plugin-client`, `@kubb/plugin-ts`, `@kubb/plugin-zod`

## Additional Context

This is a common use case for Next.js users who want to:
1. Generate type-safe API clients with Kubb
2. Use them as Server Actions (requires `"use server"` directive)
3. Organize code with `group` option for better structure

The current workaround requires platform-specific scripts (bash/perl) which is not ideal for cross-platform development.

stijnvanhulle:
When a plugin generates an index.ts with output.banner (e.g. "'use server'"),
middleware-barrel's barrel file at the same path was inheriting that banner via
FileManager.mergeFile, which always kept the existing file's banner/footer.

Fix mergeFile so the incoming file (b) takes precedence for banner/footer
("latest write wins"). Since middleware-barrel's barrel files carry no banner
or footer, they now correctly clear any banner set by a plugin-generated file
at the same path.

Also explicitly set banner/footer to undefined in makeBarrel to signal intent
and ensure correct behaviour regardless of merge order.

Closes #2265

https://claude.ai/code/session_01VHNcLfhuJ3j787YhZTRTkK


stijnvanhulle:
@copilot add changeset

PR #3226, opened 2026-05-03
feat(core): make adapter and input optional in createKubb

stijnvanhulle:
## Summary

- Makes `adapter` and `input` optional in both `Config` and `UserConfig`
- Removes the early `throw new Error('Adapter should be defined')` in `setup()`
- Wraps the adapter parsing block in `if (config.adapter) { … }` so Kubb works without a spec input
- Moves the input null-check into `inputToAdapterSource` with a clear error when an adapter is configured but no input is provided
- Updates JSDoc to document plugin-only mode (no adapter/input)

## Behaviour

| Before | After |
|---|---|
| `adapter` required — throws at startup if omitted | `adapter` optional — plugin-only mode when omitted |
| `input` required when adapter present | `input` still required at runtime when `adapter` is set (throws from `inputToAdapterSource`), optional in types |
| Dummy OAS spec needed for file-injection scripts | Plain `createKubb({ output, plugins })` works |

## Plugin-only mode

When `adapter` is omitted:
- `kubb:plugin:setup` fires normally — `injectFile` works
- Files injected via `injectFile` are written through `fsStorage` as usual
- `kubb:build:start` is **not** emitted (no `inputNode`)
- `kubb:generate:schema` / `kubb:generate:operation` are **never** emitted
- Plugins that declare generators will throw if they run without an adapter (existing `runPluginAstHooks` guard is unchanged)

## Motivation

Enables scripts that use Kubb purely for its file-management layer (e.g. `build-plugin-yaml.ts` in `kubb-labs/plugins`) without needing a dummy OpenAPI spec.

## Test plan

- [ ] Existing tests pass (`pnpm test`)
- [ ] `createKubb({ output: { path: '.' }, plugins: [myFilePlugin()] })` builds without adapter
- [ ] `createKubb({ adapter: adapterOas() })` (no input) throws `[kubb] input is required when using an adapter`
- [ ] Normal OAS builds with adapter + input are unaffected

---
_Generated by [Claude Code](https://claude.ai/code/session_01KC6q6im2qwSDF8FB3pzxKC)_

### DIFF
--- a/.changeset/make-adapter-and-input-optional.md
+++ b/.changeset/make-adapter-and-input-optional.md
@@ -1,0 +1,31 @@
+---
+"@kubb/core": minor
+---
+
+Make `adapter` and `input` optional in `Config` and `UserConfig` to support plugin-only mode.
+
+When `adapter` is omitted, Kubb skips the spec parse phase and runs in plugin-only mode:
+- `kubb:plugin:setup` fires normally — `injectFile` works as usual
+- Files injected via `injectFile` are written through storage as usual
+- `kubb:build:start` is **not** emitted (no `inputNode`)
+- `kubb:generate:schema` / `kubb:generate:operation` are never emitted
+
+This enables scripts that use Kubb purely for its file-management layer without needing a dummy OpenAPI spec.
+
+```ts
+// before — a dummy spec was required even for file-injection-only scripts
+export default defineConfig({
+  input: { path: './dummy.yaml' },
+  output: { path: '.' },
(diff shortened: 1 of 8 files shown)


stijnvanhulle:
@copilot fix typecheck

stijnvanhulle:
@copilot add changeseg

stijnvanhulle:
@copilot what else could we add/improve to reduce the code of scripts in https://github.com/kubb-labs/plugins/pull/127

PR #3249, opened 2026-05-07
feat(adapter-oas): hide parseDocument/parseFromConfig/validateDocument, add validate to adapterOas

stijnvanhulle:
$(cat <<'EOF'
## Summary

- Removes `parseDocument`, `parseFromConfig`, and `validateDocument` from the public API of `@kubb/adapter-oas` — these are implementation details that should not be exposed
- Adds an optional `validate?` method to the core `Adapter<T>` type so adapters can declare validation support
- Implements `adapter.validate(input, options?)` on the `adapterOas` instance — it parses the file at `input` then validates it, replacing the two-step dance done by the CLI
- Keeps `mergeDocuments` and `ValidateDocumentOptions` exported (still useful publicly)
- Updates the CLI `validate` runner and its test to use `adapterOas().validate()` instead of the now-hidden standalone functions

## Test plan

- [ ] `pnpm --filter @kubb/adapter-oas typecheck` passes
- [ ] `pnpm --filter @kubb/cli typecheck` passes
- [ ] `pnpm --filter @kubb/core typecheck` passes
- [ ] CLI validate test mocks `adapterOas` and asserts `validate` is called with the right args
- [ ] Internal usages (`adapter.ts`, test files, `mocks/oas.ts`) all import from `./factory.ts` directly and are unaffected

https://claude.ai/code/session_01SbAxASzL7PeeYXiJz8qFm4
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01SbAxASzL7PeeYXiJz8qFm4)_

### DIFF
--- a/packages/adapter-oas/src/adapter.ts
+++ b/packages/adapter-oas/src/adapter.ts
@@ -1,5 +1,4 @@
 import { ast, createAdapter } from '@kubb/core'
-import type { AdapterSource } from '@kubb/core'
 import { DEFAULT_PARSER_OPTIONS } from './constants.ts'
 import { applyDiscriminatorInheritance } from './discriminator.ts'
 import { parseDocument, parseFromConfig, validateDocument } from './factory.ts'
@@ -73,17 +72,8 @@ export const adapterOas = createAdapter<AdapterOas>((options) => {
     get inputNode() {
       return inputNode
     },
-    async validate(input: string | AdapterSource | Document, options) {
-      let document: Document
-
-      if (typeof input === 'string') {
-        document = await parseDocument(input)
-      } else if ('type' in input) {
-        document = await parseFromConfig(input)
-      } else {
-        document = input
-      }
-
+    async validate(input, options) {
+      const document = await parseDocument(input)
       await validateDocument(document, options)
     },
     getImports(node, resolve) {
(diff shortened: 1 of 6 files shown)


packages/adapter-oas/src/adapter.ts
stijnvanhulle, line 94: @copilot remove as we want to use adapter.validate() instead

stijnvanhulle: @copilot or we can use this.validate

stijnvanhulle: @copilot revert last commit so we keep using validate without calling this.validate in parse

packages/core/src/types.ts
```diff
@@ -130,6 +130,11 @@ export type Adapter<TOptions extends AdapterFactoryOptions = AdapterFactoryOptio
    * return `{ name, path }` for the import, or `undefined` to skip it.
    */
   getImports: (node: SchemaNode, resolve: (schemaName: string) => { name: string; path: string }) => Array<ImportNode>
+  /**
+   * Validate the document at the given path or URL.
+   * Adapters that support validation implement this method; others leave it `undefined`.
+   */
+  validate?: (input: string, options?: { throwOnError?: boolean }) => Promise<void>
```
stijnvanhulle: @copilot this should always be set

stijnvanhulle:
@copilot add changeset

PR #3285, opened 2026-05-13
perf(core,ast,agent): reduce build-time memory by reusing the Storage abstraction

stijnvanhulle:
## Why

Two retainers dominate Kubb's heap during a build:

1. **`BuildOutput.sources`** previously held every rendered file source as an in-memory `Map<string, string>` while `config.storage` (default `fsStorage()`) had already written the same bytes to disk — pure duplication.
2. **Per-`FileNode` AST graph** (`sources[].nodes: CodeNode[]`) survived the entire build, even though nothing reads it after a file is rendered.

Plus, all generated files accumulated in `FileManager` before any write happened, holding the full graph at peak.

This PR addresses the write-phase / lifetime concerns. The remaining stripe-class OOM (parsing time) is documented as a follow-up — see "What this does **not** solve" below.

## What

- **`BuildOutput.storage`** (renamed from `sources`) is now a read-through `Storage` view, scoped to the files written during the build and backed by `config.storage`. Reads are async (`await storage.getItem(path)`); the build doesn't keep a second copy of every source string in memory.
- **Streaming per-plugin write**: `FileProcessor` is invoked **after each plugin** emits its files (including barrel files added by `middleware-barrel` during that plugin's `kubb:plugin:end`). A second flush after `kubb:plugins:end` handles root barrels. Inter-plugin file merging continues to work via FileManager's path-based merge — `middleware-barrel` reads file metadata, not `source.nodes`, so disposing the AST after write is safe.
- **`source.nodes` disposal** inline in `FileProcessor.processOne` after each file is written, releasing the heavy `CodeNode` payload while preserving wrapper metadata (`sources`/`imports`/`exports` arrays + their `.name`/`.isExportable`/etc.).
- **`FileManager.dispose()`** + clears `inputNode` from `PluginDriver.dispose()` at end-of-build so the parsed adapter graph is released after `kubb:build:end`. Returned `BuildOutput.files` still references the FileNodes the caller needs.
- **`@kubb/agent`** WebSocket bridge migrated from `sources.forEach(...)` to `for (const key of await storage.getKeys()) { await storage.getItem(key) }`.

## What this does **not** solve

Profiling stripe (1385 schemas) shows the OOM happens **inside `@kubb/adapter-oas`'s `parseOas()`**, specifically in `Object.entries(schemaObjects).map(_parseSchema)` (`parser.ts:988`), before any plugin runs:

```
adapter.parse start             rss=241MB heap= 84MB
after parseFromConfig           rss=327MB heap=159MB
after validateDocument          rss=327MB heap=159MB
parseOas start                  rss=327MB heap=159MB
after getSchemas (1385 schemas) rss=327MB heap=166MB
                               ✖ OOM at 8 GB heap here ✖
```

Holding all 1385 expanded `SchemaNode` trees in a single array exceeds 8 GB for stripe's heavily circular type graph. The write-phase work in this PR cannot help here — the build dies before the plugin loop starts.

A **follow-up PR** will move per-schema parsing to a lazy, `Storage`-backed shape: `InputNode.schemas` becomes `{ names: string[], get(name): Promise<SchemaNode> }`, with `fsStorage` serializing each parsed node and a small LRU on top. Tracked separately.

## Breaking changes (beta)

- `BuildOutput.sources` → `BuildOutput.storage` (type changes from `Map<string, string>` to `Storage`).
- `kubb:generation:end` payload renamed `sources` → `storage`; callers that did `sources.get(path)` synchronously must `await storage.getItem(path)`.
- Code that read `driver.fileManager.files` after a build returned should read `BuildOutput.files` instead — the live FileManager cache is disposed at end-of-build. Two core tests were updated.

## Test plan

- [x] `pnpm typecheck` — all 11 packages
- [x] `pnpm test` — 1327 / 1328 (1 pre-existing todo)
- [x] `pnpm lint` / `pnpm format` — clean
- [x] Twitter spec (376 KB) clean A/B at HEAD vs base of branch: median 336 MB vs 342 MB (~2% improvement; small spec, not memory-bound)
- [x] Stripe heap profile captured (above) — write-phase changes correct but parser-side bottleneck remains
- [ ] kubb-labs/plugins#156: stripe E2E will still OOM until the adapter-oas follow-up; recommend keeping stripe's CI cap at 3 GB (existing) and gating it as known-failing until the follow-up lands

https://claude.ai/code/session_016WDSK4PjR3N3P9ByPbzCZ9


packages/core/src/createKubb.ts
```diff
@@ -683,9 +691,13 @@ export type BuildOutput = {
...
   /**
-   * Raw generated source, keyed by absolute file path.
+   * Read-only `Storage` view over the files written during this build.
+   * Use `await sources.getItem(path)` to read a generated file's contents on demand.
+   *
+   * Backed by `config.storage` (defaults to `fsStorage()`), so generated source bytes
+   * stay on disk instead of being held in memory for the lifetime of the build.
    */
-  sources: Map<string, string>
+  sources: Storage
```
stijnvanhulle: @copilot find better name for sources

packages/core/src/FileProcessor.ts
```diff
@@ -69,6 +69,14 @@ export class FileProcessor {
         percentage,
         total,
       })
+
+      // Release the heavy AST payload (CodeNode trees inside each SourceNode)
+      // now that the rendered string has been handed off. The FileNode shape
+      // and wrapper metadata are preserved for any consumer that still inspects
+      // the returned `BuildOutput.files`.
+      for (const source of file.sources) {
```
stijnvanhulle: @copilot do not add this as we will resolve this in a future update

packages/core/src/createKubb.ts
```diff
@@ -981,7 +1157,7 @@ async function safeBuild(setupResult: SetupResult): Promise<BuildOutput> {
         inputNode: driver.inputNode,
         getPlugin: driver.getPlugin.bind(driver),
         get files() {
-          return driver.fileManager.files
```
stijnvanhulle, line 984: @copilot keep returning the driver.fileManager.files, we will improve in a future pr

packages/core/src/createKubb.ts
```diff
@@ -571,7 +572,25 @@ export type KubbGenerationStartContext = {
 export type KubbGenerationEndContext = {
   config: Config
   files: Array<FileNode>
```
stijnvanhulle: @copilot could this be replaced by storage?

packages/core/src/createKubb.ts
```diff
@@ -1236,8 +1339,8 @@ export function createKubb(userConfig: UserConfig, options: CreateKubbOptions =
     get hooks() {
       return hooks
     },
-    get sources() {
-      return setupResult?.sources ?? new Map()
+    get storage() {
+      return setupResult?.storage ?? memoryStorage()
     },
     get driver() {
       return setupResult?.driver
```
stijnvanhulle: @copilot setupresult should always be defined do thros instead. Setup will always be called first.

packages/core/src/index.ts
```diff
@@ -12,7 +12,7 @@ export { defineParser } from './defineParser.ts'
 export { definePlugin } from './definePlugin.ts'
 export { defineResolver } from './defineResolver.ts'
 export { FileManager } from './FileManager.ts'
-export { FileProcessor } from './FileProcessor.ts'
+export { FileProcessor, type FileProcessorEvents } from './FileProcessor.ts'
```
stijnvanhulle: @copilot only expose type in types.ts

PR #3291, opened 2026-05-14
refactor: remove inputNode from Adapter interface

stijnvanhulle:
## Summary

- Remove `inputNode: InputNode | null` from the `Adapter<TOptions>` type in `createAdapter.ts`
- Remove the `get inputNode()` getter from `adapterOas` in `adapter-oas/src/adapter.ts`
- Update `createMockedAdapter` in `mocks.ts` to drop the `inputNode` option; add `inputNode?: InputNode` to `RenderGeneratorOptions` so tests can pass it directly to the generator context

## Motivation

`GeneratorContext` already exposes `inputNode: InputNode` as a guaranteed-non-null field when generators run. Storing the same (potentially large) parsed AST on the adapter as well doubled its memory footprint for no benefit. Plugins now read from `ctx.inputNode` directly.

## Test plan

- [ ] `pnpm typecheck` passes in kubb repo
- [ ] `pnpm test` passes in kubb repo
- [ ] Companion PR in kubb-labs/plugins updates all call sites

https://claude.ai/code/session_01Um1gYNswDED59mYFW2Y7m5

---
_Generated by [Claude Code](https://claude.ai/code/session_01Um1gYNswDED59mYFW2Y7m5)_

### DIFF
--- a/packages/adapter-oas/src/adapter.ts
+++ b/packages/adapter-oas/src/adapter.ts
@@ -44,10 +44,8 @@ export const adapterOas = createAdapter<AdapterOas>((options) => {
     emptySchemaType = unknownType || DEFAULT_PARSER_OPTIONS.emptySchemaType,
   } = options
 
-  // Let-binding so parse() can replace it with a simple reassignment (no clear+loop).
   let nameMapping = new Map<string, string>()
   let parsedDocument: Document | null
-  let inputNode: ast.InputNode | null
 
   return {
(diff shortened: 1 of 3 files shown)


packages/adapter-oas/src/adapter.ts
stijnvanhulle: @copilot do we still need this scoped let? Can we use a const as part of parser?

PR #3290, opened 2026-05-14
Adapter streaming with AsyncIterable

stijnvanhulle:
$(cat <<'EOF'
## Summary

- Adds `docs/architecture/adr/0005_stream_adapter.md` documenting the design decision to add an optional `stream()` method to the `Adapter` interface
- No code changes — this is a planning/architecture document only

## What the ADR covers

The ADR describes adding an `AsyncGenerator<AdapterStreamEvent>`-based `stream()` method to the adapter interface so the build orchestrator can process schema and operation nodes one at a time, allowing the GC to reclaim each parsed AST node after all plugins have touched it.

Key points documented:
- What stays in memory regardless (raw OAS document — unavoidable due to Redocly $ref resolution)
- What streaming eliminates (parsed `SchemaNode[]` / `OperationNode[]` heap)
- Known limitations (allowedSchemaNames optimisation skipped, driver.inputNode.schemas empty during streaming)
- Considered and rejected alternatives

## Test plan

- [ ] Review ADR content for accuracy and completeness
- [ ] Confirm file is at `docs/architecture/adr/0005_stream_adapter.md`

https://claude.ai/code/session_0157cixdyQFAyr46mG7abGxZ
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_0157cixdyQFAyr46mG7abGxZ)_

PR #3292, opened 2026-05-15
fix(streaming): fan-out single-pass + lazy discriminator patching — resolves PR #3290 issues

stijnvanhulle:
## Summary

Fixes two bugs in the streaming adapter implementation from PR #3290, then simplifies the operations API.

### Bug 1: N×parse-time overhead for multi-plugin builds (+263% time)

`runPluginAstHooks` was called once per plugin in streaming mode. Each call created a fresh `for await` iterator that re-parsed all schemas from scratch. For a 5-plugin bunq build: 5 × 641 ms = **3,205 ms of redundant parsing**.

**Fix: `runStreamingFanOut` in `packages/core/src/createKubb.ts`**

New function iterates `inputStreamNode.schemas` and `.operations` exactly **once**, distributing each node to all plugins simultaneously. Build cost is now 1×parse regardless of plugin count.

```
Before: plugin₁ → full stream pass → plugin₂ → full stream pass → ... → pluginN → full stream pass
After:  single stream pass → fan schema/op to [plugin₁, plugin₂, ..., pluginN]
```

Plugin lifecycle events (`plugin:start`/`end`, timings, error handling per plugin) are preserved.

### Bug 2: `discriminator: 'inherit'` buffered all schemas (defeated memory savings)

In `adapter.ts`, the `discriminator === 'inherit'` branch parsed ALL schemas into an array upfront before yielding any — the exact pattern streaming was designed to avoid.

**Fix: lazy discriminator patching in `packages/adapter-oas/src/adapter.ts`**

Pre-scan raw OAS schema objects to find discriminator-parent schemas (those with `oneOf`/`anyOf` + `discriminator.propertyName`) — typically a small subset. Parse only those to AST and build the child map. All other schemas are yielded lazily with the patch applied inline per yield.

**Refactor `packages/adapter-oas/src/discriminator.ts`:** Extracted `buildDiscriminatorChildMap(schemas)` and `patchDiscriminatorNode(node, entry)` as exported helpers.

## Operations API Simplification

`gen.operations()` now always receives `OperationNode[]` (plain array) in both the batch and streaming paths. During the streaming fan-out, operations are accumulated into an array while processing per-node `gen.operation()` calls, then the collected array is passed to `gen.operations()`.

This removes the `AsyncIterable<OperationNode> | Array<OperationNode>` union from the generator type — plugins receive a simple array as before. `collectOperations` has been deleted.

## Code Cleanup

- `AdapterCache` simplified to a `Storage` alias — `config.storage` used directly via a scoped prefix wrapper
- `cachedSchemaObjects` closure variable eliminates redundant `getSchemas()` call between `count()` and `stream()`
- `STREAM_SCHEMA_THRESHOLD` and `STREAM_FLUSH_EVERY` consolidated into `constants.ts`
- `collectOperations` removed (no longer needed)

## Tests

- All `adapter.test.ts` tests pass (count, stream, disk cache)
- All `core` tests pass
- TypeScript: zero errors in `packages/core` and `packages/adapter-oas`

https://claude.ai/code/session_01JYiUeNJ39LBChYBdBziDtb


stijnvanhulle:
@copilot resolve the merge conflicts in this pull request

packages/core/src/constants.ts
```diff
@@ -24,6 +24,16 @@ export const DEFAULT_BANNER = 'simple' as const
...
 
+/**
+ * Schema count above which the adapter's `stream()` path is used instead of `parse()`.
+ */
+export const STREAM_SCHEMA_THRESHOLD = 100
+
+/**
+ * In streaming mode, flush generated files to disk every N schemas to bound in-memory file buffers.
+ */
+export const FLUSH_EVERY = 50
```
stijnvanhulle: @copilot add stream prefix

packages/core/src/createAdapter.ts
```diff
@@ -3,18 +3,12 @@ import type { ImportNode, InputNode, InputStreamNode, SchemaNode } from '@kubb/a
...
+ * Adapters call `cache.getItem(hash)` / `cache.setItem(hash, json)` directly — key scoping
+ * is handled by the core before the source reaches the adapter.
  */
-export type AdapterCache = {
-  /** Storage backend to use for reading and writing cache entries. */
-  storage: Storage
-  /** Absolute path to the cache directory, e.g. `/project/.kubb/.cache`. */
-  dir: string
-}
+export type AdapterCache = Storage
```
stijnvanhulle: @copilot no need for new type, use Storage directly

PR #3305, opened 2026-05-16
perf: convert array-building loops to generator functions

stijnvanhulle:
## Summary

- **`fsStorage`**: `walk()` converted to an async generator that yields keys as directories are read, eliminating the closure-mutating `keys` array and enabling future callers to stream results
- **`transformers`**: new `mergeAdjacentObjectsLazy()` generator streams merged object nodes without a reduce accumulator; `mergeAdjacentObjects()` delegates to it. `mergeAdjacentObjectsLazy` is exported from `@kubb/ast`
- **`parser` (operations)**: replaces `flatMap(...).filter()` with a generator that skips null operations in one pass — no intermediate null-padded array, no second filter sweep
- **`parser` (content types)**: replaces `flatMap` with empty-array `return []` per miss with a `continue`-based generator, eliminating a temporary `[]` allocation for every unsupported content type per operation
- **`parser` (allOf members)**: uses `mergeAdjacentObjectsLazy` directly in the spread so the two slice groups are streamed into `members` without creating intermediate result arrays

## Test plan

- [ ] `pnpm --filter @kubb/ast test run` — 301 tests pass
- [ ] `pnpm --filter @kubb/adapter-oas test run` — 352 tests pass (2 pre-existing failures on `#mocks` virtual module unrelated to these changes)
- [ ] `pnpm --filter @kubb/ast typecheck` — clean
- [ ] `pnpm --filter @kubb/adapter-oas typecheck` — clean
- [ ] `pnpm --filter @kubb/core typecheck` — clean

https://claude.ai/code/session_01YWZT4xigczvoTuR8HQAYEq

---
_Generated by [Claude Code](https://claude.ai/code/session_01YWZT4xigczvoTuR8HQAYEq)_


packages/ast/src/transformers.ts
```diff
@@ -73,25 +73,30 @@ export function setDiscriminatorEnum({
  * ])
  * ```
  */
-export function mergeAdjacentObjects(members: Array<SchemaNode>): Array<SchemaNode> {
-  return members.reduce<Array<SchemaNode>>((acc, member) => {
+export function* mergeAdjacentObjectsLazy(members: Iterable<SchemaNode>): Generator<SchemaNode, void, undefined> {
+  let pending: SchemaNode | undefined
```
stijnvanhulle: @copilot can we do this without pending state?

packages/ast/src/transformers.ts
```diff
@@ -73,27 +73,29 @@ export function setDiscriminatorEnum({
  * ])
  * ```
  */
-export function mergeAdjacentObjects(members: Array<SchemaNode>): Array<SchemaNode> {
-  return members.reduce<Array<SchemaNode>>((acc, member) => {
+export function mergeAdjacentObjectsLazy(members: Iterable<SchemaNode>): Array<SchemaNode> {
```
stijnvanhulle: @copilot use function generator pattern with yield

packages/ast/src/transformers.ts
```diff
@@ -73,25 +73,36 @@ export function setDiscriminatorEnum({
...
+      continue
     }
 
-    acc.push(member)
-    return acc
-  }, [])
+    let merged = objectNode
+    i++
+
+    while (i < items.length) {
```
stijnvanhulle: @copilot dont use while of for loops, keept simple aa it was before with claude and pending state

packages/adapter-oas/src/parser.ts
```diff
@@ -886,17 +886,16 @@ function createSchemaParser(ctx: OasParserContext) {
 
     const requestBodyMeta = getRequestBodyMeta(operation)
 
-    const content = allContentTypes.flatMap((ct) => {
+    const content: Array<{ contentType: string; schema: ast.SchemaNode; keysToOmit: Array<string> | undefined }> = []
```
stijnvanhulle: @copilot why not flatMap, is there a difference in memory/speed?

packages/ast/src/transformers.ts
stijnvanhulle: @copilot find better name for pending

PR #3310, opened 2026-05-17
feat(core): add FileProcessor.runStream() and use it in flushPendingFiles

stijnvanhulle:
## Summary

- Adds `FileProcessor.runStream()` — an async generator that yields one `ParsedFile` at a time instead of emitting batched events
- Replaces the three `fileProcessor.events.on(start/update/end)` listeners in `safeBuild()` with a single `for await...of fileProcessor.runStream()` loop
- Exports `ParsedFile` type from `@kubb/core`

## Why

The existing `run()` method uses `Promise.all` + `pLimit(16)` + `AsyncEventEmitter` dispatch per file. For large builds this means:
- Up to 16 parsed strings held in memory simultaneously (event-handler closure retention)
- O(N) promise allocation upfront before any file reaches storage
- Per-file emitter dispatch overhead on every write

`runStream()` eliminates all three: one file is parsed, yielded, written, and GC-eligible before the next starts. The `for await` loop is the scheduler — no pLimit queue, no emitter.

## Performance (measured against petStore + 4 plugins, `memoryStorage`)

| Metric | `run()` parallel | `runStream()` | Delta |
|---|---|---|---|
| Throughput (ops/s) | 257,431 | 297,416 | **+15.5%** |
| Mean latency (ms) | 0.0039 | 0.0034 | **−13%** |
| First-write latency (ops/s) | 234,442 | 280,272 | **+19.6%** |
| Peak-heap overhead (ops/s) | 103,829 | 115,367 | **+11.1%** |
| p999 tail latency (ms) | 0.0575 | 0.0374 | **−35%** |

Full benchmark suite and report in `kubb-labs/plugins` PR `claude/validate-stream-adapter-RUfiC`.

## Backwards compatibility

`run()` is unchanged. `runStream()` is additive. No breaking changes.

## Changes

- `packages/core/src/FileProcessor.ts` — adds `ParsedFile` type + `runStream()` method
- `packages/core/src/createKubb.ts` — `flushPendingFiles()` uses `runStream()`, removes event-listener boilerplate
- `packages/core/src/index.ts` — re-exports `ParsedFile`

---
_Generated by [Claude Code](https://claude.ai/code/session_01JeSDaMTRwnBVCvjyvkwQxE)_


packages/core/src/index.ts
```diff
@@ -13,6 +13,7 @@ export { definePlugin } from './definePlugin.ts'
 export { defineResolver } from './defineResolver.ts'
 export { FileManager } from './FileManager.ts'
 export { FileProcessor } from './FileProcessor.ts'
+export type { ParsedFile } from './FileProcessor.ts'
```
stijnvanhulle: @copilot should be pary of types.ts

PR #3315, opened 2026-05-18
feat: keep format info in schema

Ericlm:
## 🎯 Changes

Add a new property `format` to `SchemaNodeBase` which contains the original `format` value. This will allow downstream plugins to consume the generated format.

## ✅ Checklist

## 🚀 Release Impact

Related https://github.com/kubb-labs/plugins/issues/167

### DIFF
--- a/.changeset/cool-jokes-pump.md
+++ b/.changeset/cool-jokes-pump.md
@@ -1,0 +1,5 @@
+---
+"@kubb/adapter-oas": patch
+---
+
+Add format property to SchemaNodeBase.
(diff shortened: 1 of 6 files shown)


stijnvanhulle:
@Ericlm just thinking about the future, will other schemas like graphql and such also have format field? Or could we use another name

Ericlm:
@stijnvanhulle there is no "format" property in GraphQL, but I don't know how we could name it 🤔

stijnvanhulle:
`format` is fine as a name. Both AsyncAPI and Arazzo inherit JSON Schema's `format` keyword with the same semantics as OAS, so there's no conflict. GraphQL and Prisma don't have a `format` concept at all — they'd consume this field as an input signal (e.g. `format: 'uuid'` → `@db.Uuid` in Prisma) rather than needing to write their own value into it.

stijnvanhulle (approved):

PR #3475, opened 2026-06-04
fix(plugin-ts): preserve object-array types that contain nested enums

joaomlneto:
## Summary

With `enumType: 'asConst'` or `'asPascalConst'` (asConst is the default), an array whose items are an **object** containing an enum property is generated as a bare enum array — the object shape is silently dropped.

In `packages/plugin-ts/src/components/Type.tsx`, `isDirectEnum` matched **any** array that had `items`:

```ts
const isDirectEnum = schema.type === 'array' && schema.items !== undefined
```

When the tree contained any enum (found via `deepSearch`, which descends into object properties), the correctly-parsed object type was replaced by `enumSchemas[0]` wrapped in an array.

## Reproduction

```yaml
ArrayOfObject:
  type: array
  items:
    type: object
    required: [id, status]
    properties:
      id: { type: integer }
      status: { type: string, enum: [active, inactive] }
```

**Before** (`enumType: 'asConst'`):

```ts
export type ArrayOfObject = ArrayOfObjectStatusEnumKey[] // ❌ id/status object lost
```

**After:**

```ts
export type ArrayOfObject = {
  id: number
  status: ArrayOfObjectStatusEnumKey
}[]
```

Real-world impact: against the EVE Online ESI spec this mis-generated ~45 response types (character/corporation assets, contacts, mail, market orders, …).

## Fix

Only treat the schema as a direct enum array when the array's **items** are themselves an enum, not when the items are an object that merely contains an enum property:

```ts
const isDirectEnum =
  schema.type === 'array' &&
  typeof schema.items === 'object' &&
  schema.items !== null &&
  !Array.isArray(schema.items) &&
  'enum' in schema.items
```

Inline-enum arrays (`items: { type: string, enum: [...] }`) and `$ref`-to-enum arrays are unaffected.

## Test

Added an `enum.ArrayOfObject` schema to `mocks/enums.yaml` and an `EnumArrayOfObject` case to `typeGenerator.test.tsx`; the snapshot shows the object shape is preserved. Full suite (`pnpm --filter @kubb/plugin-ts test`) passes.

Targets the `v4` branch (4.37.9). Happy to adjust naming/approach.


stijnvanhulle:
@all-contributors please add @joaomlneto for code

PR #3682, opened 2026-06-28
fix(adapter-oas): resolve collision-renamed component imports

stijnvanhulle:
## 🎯 Changes

When two components share a name across sections or by case (for example `#/components/schemas/Order` and `#/components/requestBodies/Order`, or `Variant`/`variant`), `getSchemas` disambiguates them into distinct emitted files (`OrderSchema`, `OrderRequest`, `Variant2`) and records each rename in a `nameMapping` keyed by the **full `$ref` path** (already exposed on `adapter.options.nameMapping`).

`getImports` looked the map up by the **short** ref name (`extractRefName(...)`), so it always missed and emitted an import for the un-disambiguated name — a dangling import that did not match the file actually generated (`TS2307` in downstream output).

`getImports` now resolves the collision-renamed name via `nameMapping.get(schemaRef.ref)` (the full `$ref` path) before falling back to the short ref name, so the import targets the generated file. One-line root-cause fix; no new public API.

This is the import half of a coordinated fix. The plugin printers (plugin-ts / plugin-zod / plugin-faker) read the same `adapter.options.nameMapping` to resolve the type/schema **reference** name; that side lives in the paired plugins-repo PR ([kubb-labs/plugins#579](https://github.com/kubb-labs/plugins/pull/579)).

Verified end-to-end with a linked build against the `caseSensitivity` and `digitalocean` fixtures: the previously dangling `import { Order } from './Order.ts'` now resolves to `OrderSchema`, taking the strict-typecheck collision errors (240 across the two specs) to 0.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test` (adapter-oas suite green, new collision test passing).

## 🚀 Release Impact

https://claude.ai/code/session_01CAfw3gfcB2XXbCuuQVrvLB

### DIFF
--- a/.changeset/adapter-oas-name-collision-mapping.md
+++ b/.changeset/adapter-oas-name-collision-mapping.md
@@ -1,0 +1,9 @@
+---
+'@kubb/adapter-oas': patch
+---
+
+Resolve imports to the renamed target when a component name collides.
+
+When two components share a name across sections or by case (e.g. `#/components/schemas/Order` and `#/components/requestBodies/Order`), `getSchemas` disambiguates them (`OrderSchema`, `OrderRequest`) and records the rename in a `nameMapping` keyed by the full `$ref` path. `getImports` looked the map up by the short ref name, always missed, and emitted an import for the un-disambiguated name — a dangling import that did not match the generated file.
+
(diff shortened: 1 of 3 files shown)


stijnvanhulle:
Closing as a duplicate of #3681, which is already merged to `main` and applies the identical `getImports` fix — resolving collision-renamed schema refs through `nameMapping.get(schemaRef.ref)` — along with an equivalent collision test. Nothing unique remains here.

The plugin-side counterpart (the `ref()` type-reference fix in plugin-ts/plugin-zod/plugin-faker) lives in kubb-labs/plugins#579 and remains open.

---

PR #3715, opened 2026-07-05
refactor(core): rename resolver core namespace to default, add top-level name/file

stijnvanhulle:
## 🎯 Changes

Replaces the stringly-typed `default(name, type)` naming primitive on the plugin `Resolver` with a `default` namespace of dedicated helpers, and injects two top-level entries, `name` and `file`, that every generator calls directly.

The old `default(name, type?: 'file' | 'function' | 'type' | 'const')` was ambiguous: calling it without a type silently meant camelCase, and some plugins returned different strings for `default(name)` versus `default(name, 'function')`. The built-in machinery now lives under `resolver.default`:

- `default.name(name)` — the built-in camelCase identifier casing
- `default.file(params, context)` — the `FileNode` builder; base-name casing comes from `params.resolveName` (default `toFilePath`), so the removed `default.fileName` primitive is gone
- `default.options` / `default.path` / `default.banner` / `default.footer` — the other built-ins, unchanged in behavior

Two top-level members are injected onto every resolver — `resolver.name(name)` and `resolver.file(params, context)` — each delegating to its `default` counterpart unless a plugin overrides it directly (flat, no `default:` block needed for the common case):

```ts
export const resolverTs = defineResolver(() => ({
  pluginName: 'plugin-ts',
  name(name) {
    return ensureValidVarName(pascalCase(name))
  },
  file(params, context) {
    return this.default.file({ ...params, resolveName: (name) => toFilePath(name, pascalCase) }, context)
  },
  response: {
    status(node, code) {
      return this.name(`${node.operationId} Status ${code}`)
    },
  },
}))
```

A plugin's own naming methods are grouped into namespaces alongside `default` (`response`, `param`, `query`, ...) and reach the shared helpers through `this` — `this.name(...)`, `this.default.*`, `this.file(...)` — since the resolver root is what `this` resolves to inside a namespace method (`bindNamespaces`, a single shallow pass). Overrides merge through `mergeResolver(base, override)`, which accepts a deep partial, so an override can name a single top-level entry, a single namespace method, or a single `default` helper without restating the rest.

This is a breaking change to the `@kubb/core` resolver API: `Resolver.name` (the string identifier field) is removed, freeing `name` for the method; `default.fileName` is removed in favor of `file`'s `resolveName` param. Updates the core call sites (`defaultResolveFile`, the driver's option filtering, the barrel plugin's banner/footer) and reconciles the in-flight resolver changesets.

## ✅ Checklist

## 🚀 Release Impact

https://claude.ai/code/session_01WnjrGJLD5PpnVdeJhqx5en

---
_Generated by [Claude Code](https://claude.ai/code/session_01WnjrGJLD5PpnVdeJhqx5en)_


stijnvanhulle:
Is streaming actually making kubb faster? No, not on a local disk. I measured it with a synthetic that mimics the real loop (CPU work per node, enqueue, yield): 2000 files, streaming      
  writes 346-380ms vs write-everything-at-drain 330-349ms. Batch-at-end consistently wins by a few percent. The reason is structural: parsing (formatting) runs on the same thread as         
  generation, so the only thing that can genuinely overlap is the raw fs.writeFile calls, and on an SSD those are absorbed by libuv's threadpool in microseconds. Streaming pays per-batch    
  hook emissions in exchange for hiding IO that was already nearly free.                                                                                                                      
                                                                                                                                                                                              
  What streaming does still buy: live progress in the CLI during long builds (files appear on disk as they generate instead of all at the end), and a real win if someone plugs in a slow     
  Storage (network fs, remote storage). Since the new design makes streaming cost essentially nothing on the driver side, keeping it is fine, but if you ever want to simplify further,       
  dropping the mid-build writes and writing once at drain() would be both simpler and marginally faster for the default fs case.

PR #3767, opened 2026-07-10
fix(core): preset name casing regression

Ericlm:
## ✅ Checklist

## 🚀 Release Impact


Ericlm:
@stijnvanhulle I made a failing test in this repo because it seems that it is causing all generated types with plugin-ts to have camelCase instead of PascalCase. Could you provide me some guidance?

packages/core/src/createResolver.test.ts
```diff
@@ -247,6 +247,26 @@ describe('createResolver', () => {
...
+      pluginName: 'test',
+      name(name) {
+        return name.toUpperCase()
+      },
+      greet: (name: string) => name,
+      farewell: (name: string) => name,
+    })
+    const customized = Resolver.merge(preset, {
+      name(name) {
+        return this.default.name(name)
```
stijnvanhulle, line 261: this.default.name never gets changes this is the core logic of a resolver. This means that 

```name(name) {
        return name.toUpperCase()
      },
``` 

has not impact on `this.default.name`, would it make sense to repeat  return name.toUpperCase() again?

Ericlm: I believe I understand what you mean now; my concern was that `this.default.name(name)` was giving me camelCase instead of PascalCase, while I was resolving typescript types.

This is a documentation issue in this case:
https://kubb.dev/docs/5.x/guide/going-further/resolvers#rename-identifiers

Would you like me to update the doc?

stijnvanhulle: ooh jeah indeed, I made that change so the docs are not correct anymore and yes please if you would like to do that. THanks :)

PR #3809, opened 2026-07-16
Best practices discussion

devinrhode2:
For discussion


stijnvanhulle:
@devinrhode2 in v5 we will use the docs that are inside their own repo so best to check those docs and update there: [docs](https://github.com/kubb-labs/docs)

Issue #3863, opened 2026-08-13
perf(core): generation spends its time on redundant filesystem syscalls


stijnvanhulle:
CPU-profiled a 2000-file generation (`--cpu-prof`, `output.format: false`, cold plus warm build) to find out where time actually goes before optimizing anything.

## What the profile says

Generation is I/O-bound, not compute-bound. Of 1375 ms sampled:

| | share |
| --- | --- |
| `openFileHandle` | 14.4% |
| `close` | 9.4% |
| `mkdir` | 4.8% |
| `fstat` | 4.7% |
| `read` | 4.0% |
| `open` | 3.6% |
| `writeBuffer` | 3.4% |
| garbage collector | 5.4% |
| `combineSources` | 2.5% |
| `createFile` | 1.8% |

Filesystem syscalls are roughly 45%. Kubb's own JS is about 4%. Optimizing generators or the AST walk would be optimizing the wrong few percent; the wins are in syscall count and allocation pressure.

## 1. `write()` calls `mkdir` for every file

`internals/utils/src/fs.ts` runs `mkdir(dirname(resolved), { recursive: true })` before every write. A 2000-file spec writing into one directory makes 2000 of those, 1999 of which do nothing.

Measured directly:

```
2000 redundant mkdir : 262ms
2000 with a Set guard:   1ms
```

Against a ~766 ms cold build for the same spec, this is the largest single item. Fix is a per-process set of directories already ensured, or attempt the write first and only `mkdir` on `ENOENT`. Contained to one function.

## 2. The unchanged-content check allocates a copy of every file

`matchesOnDisk` calls `disk.trimEnd()`, which allocates a full copy of each file's content only to compare it and throw it away. `disk.startsWith(content)` plus a whitespace check on the short remainder answers the same question without the copy. At 2000 files that is roughly 10 MB of transient garbage removed, against 5.4% of the profile spent in GC.

Smaller than the first item and worth measuring rather than assuming.

## 3. Two concurrency limiters at the same value

`FileManager` bounds writes at `FILE_CONCURRENCY` through `inParallel`, and `fsStorage.writeItem` wraps every write in its own `createLimiter(50)`. Called through `FileManager` the inner limiter can never engage, so each write pays for a promise-queue hop it does not need. The profile shows `promiseInitHook` at 1.5% and `trackPromise` at 1.2% of async bookkeeping.

`fsStorage` is public API and a direct caller still needs a bound, so this is a design decision rather than a deletion.

[...]

PR #3871, opened 2026-08-13
perf(adapter-oas): load the OpenAPI validator only when validation runs

stijnvanhulle:
## 🎯 Changes

Also under #3863, independent of #3868 and #3870 (different package).

`load/normalize.ts` imported `compileErrors` and `validate` from `@readme/openapi-parser` at the top of the file. Both are used only inside `validateDocument`. `adapter.ts` imports that module, so importing `@kubb/adapter-oas` at all pulls the dependency in, and every `kubb.config.ts` does that.

`adapter.ts:200` is `if (validate) await validateDocument(document)`, so a project running `adapterOas({ validate: false })` paid for the heaviest dependency in the package and then never called into it.

The import now sits inside `validateDocument`.

### Measured

`@readme/openapi-parser` alone: 93.1 / 104.9 / 94.9 ms.

Importing the built adapter, alternating between the two builds:

| | main | this branch |
| --- | --- | --- |
| round 1 | 188.8 ms | 97.2 ms |
| round 2 | 138.3 ms | 76.2 ms |
| round 3 | 143.9 ms | 88.4 ms |

Median **143.9 ms → 88.4 ms**, about -39%.

Worth being precise about who gets what. This is an outright saving for `validate: false`, and for anything that loads the adapter without parsing a document. With the default `validate: true` it is a **deferral, not a saving** — the same work happens inside `parse()` instead. It still gets the dependency off the config-load path.

### ESM and CJS

tsdown leaves it as a real dynamic import in both outputs rather than downleveling to `require()`, and both were exercised against the built bundles (`parse()` with `validate: true`, plus an invalid document with `throwOnError: true` so `validate` and `compileErrors` both run). Details in [this comment](https://github.com/kubb-labs/kubb/pull/3871#issuecomment-5291499282).

One consequence worth a decision before merge: a native `await import()` inside `dist/index.cjs` resolves the `import` condition, so CJS consumers now load the ESM copy of `@readme/openapi-parser` where the static `require` gave them the CJS one. A CJS host that also requires the package directly ends up with two copies, failing `instanceof` across the boundary and keeping separate ajv caches. Nothing in this repo does that. `createRequire` in the CJS path would avoid it at the cost of a format-specific branch.

### Two things left undone

`yaml` (38.2 ms), `@scalar/openapi-upgrader` (11.0 ms), and `api-ref-bundler` (7.0 ms) are also used only inside parse-time functions and stay eager, so about 56 ms is still on the config-load path.

Nothing asserts the dependency stays off that path, so a future static import would reintroduce the cost with all 513 tests still green. A guard would need a child process loading the built entry and checking the module is absent.

## ✅ Checklist

`pnpm test` for `packages/adapter-oas` (513 passing), `pnpm lint`, `oxfmt`, and typecheck are clean.

## 🚀 Release Impact

### DIFF
--- a/.changeset/adapter-oas-lazy-validator.md
+++ b/.changeset/adapter-oas-lazy-validator.md
@@ -1,0 +1,9 @@
+---
+'@kubb/adapter-oas': patch
+---
+
+Load the OpenAPI validator only when validation runs.
+
+`@readme/openapi-parser` is the heaviest dependency in the package, and it was imported at the top of the module that holds `validateDocument`. Importing `@kubb/adapter-oas` pulled it in, which every `kubb.config.ts` does, so the cost landed on runs that never validate anything.
+
+It is now loaded inside `validateDocument`. Importing the adapter drops from 144ms to 88ms, medians of three runs. Setting `validate: false` saves that outright, and so does any tool that loads the adapter without parsing a document. With validation on, which is the default, the same work moves into the parse step instead of disappearing.
--- a/.changeset/adapter-oas-lazy-validator.md
+++ b/.changeset/adapter-oas-lazy-validator.md
@@ -4,6 +4,6 @@
 
 Load the OpenAPI validator only when validation runs.
 
-`@readme/openapi-parser` is the heaviest dependency in the package, and it was imported at the top of the module that holds `validateDocument`. Importing `@kubb/adapter-oas` pulled it in, which every `kubb.config.ts` does, so the cost landed on runs that never validate anything.
+`@readme/openapi-parser` is used by `validateDocument` alone, but it was imported at the top of the file, so every config importing `@kubb/adapter-oas` paid for it even with `validate: false`.
 
-It is now loaded inside `validateDocument`. Importing the adapter drops from 144ms to 88ms, medians of three runs. Setting `validate: false` saves that outright, and so does any tool that loads the adapter without parsing a document. With validation on, which is the default, the same work moves into the parse step instead of disappearing.
+Importing the adapter drops from 144ms to 88ms. With validation on, the cost moves into the parse step rather than going away.
--- a/packages/adapter-oas/src/load/normalize.ts
+++ b/packages/adapter-oas/src/load/normalize.ts
@@ -1,7 +1,6 @@
 import path from 'node:path'
 import { Diagnostics } from '@kubb/core'
 import type { AdapterSource } from '@kubb/core'
-import { compileErrors, validate } from '@readme/openapi-parser'
 import { upgrade } from '@scalar/openapi-upgrader'
 import { bundle } from 'api-ref-bundler'
 import { parse } from 'yaml'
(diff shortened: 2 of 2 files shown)


stijnvanhulle:
@claude can this work in esm and cjs?

stijnvanhulle:
Yes, in both. I checked it rather than assuming, because the answer depends on what tsdown does with the `await import(...)`.

tsdown leaves it as a real dynamic import in each output rather than downleveling it to `require()`:

```
dist/index.cjs:531  const { compileErrors, validate } = await import("@readme/openapi-parser");
dist/index.js:508   const { compileErrors, validate } = await import("@readme/openapi-parser");
```

That is the case that matters, since a rewrite to `require()` would be the thing that breaks. `@readme/openapi-parser` is a dual package (`exports` has both `require` and `import`), and Node supports `import()` from CJS regardless, so neither output is loading something it cannot.

Ran both against the built bundles:

```
ESM parse+validate ok, schemas: 13
ESM validate reported: OpenAPI schema validation failed.
---
CJS parse+validate ok, schemas: 13
CJS validate reported: OpenAPI schema validation failed.
```

The first line of each runs `parse()` with `validate: true` on `mocks/petStore.yaml`. The import sits outside the `try`, so a format where it failed to resolve would throw here instead of parsing. The second passes a deliberately invalid document with `throwOnError: true`, which exercises `validate` and `compileErrors` together, so both bindings are real in both formats.

One thing this does change: `validateDocument` was already `async`, so no call signature moves, but the module now loads on first validation rather than at import. A consumer that wants the old eager behavior can call `adapter.validate(...)` once up front.

---

stijnvanhulle:
Following up on my own answer, because a review of this PR turned up a consequence I did not mention and it belongs with the ESM/CJS question.

Both formats work, which is what I verified. What I left out is *which copy* the CJS build now loads. A native `await import()` inside `dist/index.cjs` resolves the `import` condition, so CJS consumers get the ESM build of `@readme/openapi-parser` where the static `require` previously gave them the CJS one. Confirmed by loading `dist/index.cjs`, running a validation, and finding no `@readme/openapi-parser` entry in `require.cache` (only its CJS transitive `@readme/better-ajv-errors`).

That matters for a CJS host that also does `require('@readme/openapi-parser')` itself: two copies live in the process, `instanceof ResolverError` and friends are false across the boundary, and the two copies keep separate ajv caches. Nothing in this repo does that, and it does not affect ESM consumers at all, but it is a real behavior change rather than a pure deferral and you should have it before merging.

If you would rather not take that, the alternative is `createRequire` in the CJS path so both formats keep loading the copy they load today, at the cost of a format-specific branch in the source.

Two other things the review raised on this PR, neither of which I have changed:

`yaml` (38.2 ms), `@scalar/openapi-upgrader` (11.0 ms), and `api-ref-bundler` (7.0 ms) are also used only inside parse-time functions and stay eager, so roughly 56 ms is still on the config-load path. The same one-line treatment would apply to them.

Nothing asserts the dependency stays off the import path, so a future edit adding the static import back reintroduces the cost silently with all 513 tests still green. A guard that loads the built entry in a child process and checks the parser is absent would pin it.

---

PR #3883, opened 2026-08-15
Fix write() dedup defeated by fsPlugin's raw writes and output.format's post-write pass

### LINKED ISSUE #3859: `output.format` runs after writing, so `write()`'s unchanged-content check never matches
**Package:** `@kubb/core` 4.39.2 (Node 24.15.0, macOS, formatter `oxfmt`)

## Summary

`write()` skips a file whose content is unchanged, but the formatter runs as a separate pass **after** all files are written. The formatter's output is not byte-identical to what `write()` compared against, so on the next build the check fails for **every file** and Kubb rewrites the entire output — even when the input document has not changed at all.

The net effect is that the unchanged-content optimisation in `write()` never takes effect for anyone who configures `output.format`.

## Where it comes from

`write()` compares the file on disk against the trimmed content it is about to write ([`src/utils/write.ts`], shipped at `dist/index.js:420-434`):

```js
async function write(path, data, options = {}) {
  const trimmed = data.trim();
  ...
  try {
    if (await readFile(resolved, { encoding: "utf-8" }) === trimmed) return void 0;
  } catch {}
  await mkdir(dirname(resolved), { recursive: true });
  await writeFile(resolved, trimmed, { encoding: "utf-8" });
```

`data.trim()` removes the trailing newline. The formatters are then invoked as CLI passes over the output directory (`dist/index.js:717-740`):

```js
const formatters = {
  prettier: { command: "prettier", args: (outputPath) => ["--ignore-unknown", "--write", outputPath] },
  biome:    { command: "biome",    args: (outputPath) => ["format", "--write", outputPath] },
  oxfmt:    { command: "oxfmt",    args: (outputPath) => [outputPath] },
};
```

Every one of those formatters writes a trailing newline (and may reflow quotes, wrapping and import order). So the bytes on disk after build N are never what `write()` produces in build N+1, and the equality check cannot succeed.

## Reproduction

1. Any project with `output.format: 'oxfmt'` (or `'prettier'`/`'biome'`).
2. Run the build twice with no change to the input document.
3. Every generated file has a new mtime after the second run.

Trailing bytes of the same file, before and after the format pass:

```
written by write():   3b 0a 7d 3b        ;\n};
after the formatter:  0a 7d 3b 0a        \n};\n
```

## Impact

Measured on a 397-file client (one OpenAPI document, `plugin-ts` + `plugin-zod` + `plugin-client` + `plugin-react-query` + `plugin-faker` + `plugin-msw`):

| run | `output.format: 'oxfmt'` | `output.format: false` |
| --- | --- | --- |
| rebuild, no input change | **397** files rewritten | **15** |
| rebuild, one property added to one DTO | **397** | **18** |

The 15 in the unformatted column are barrels and a few files written twice within one run; the rest of the tree is correctly skipped, which shows the check works exactly as intended once the formatter is out of the way.

This matters most in watch/dev setups. Rewriting the whole output on every rebuild makes file watchers (chokidar, `ts-checker-rspack-plugin`, bundler dev servers) re-run over hundreds of unchanged modules, and on Windows a watcher re-establishing a per-file watch on a file being rewritten can fail with `EBUSY`. We currently have to disable formatting in our watch mode to keep this usable.

## Suggested fix

Format the content **before** it reaches `write()`, rather than as a post-pass over the output directory. Then the bytes compared in build N+1 are the bytes written in build N, and the existing check starts doing its job.

That is a bigger change for `prettier`/`biome`/`oxfmt` as CLIs, since they are invoked per path today. Two smaller options that would also fix it:

1. **Normalise the trailing newline.** Write `` `${trimmed}\n` `` and compare against the same, so at least the newline-only difference stops forcing a rewrite. This alone would not help formatters that also reflow code, but it is a one-line change and removes the most common cause.
2. **Compare post-format content.** Keep a manifest of the content hash after the format pass and consult it before writing, so a file the formatter already normalised is recognised as unchanged.

Happy to open a PR for whichever direction you prefer.

### LINKED ISSUE #3867: Barrel files are rewritten on every build even when their content is unchanged
**Packages:** `@kubb/core`, `@kubb/plugin-oas`, `@kubb/plugin-ts` — all `4.39.2` (Node 24.15.0, macOS)

## Summary

`write()` skips a file whose content is unchanged, and that works correctly for generated modules. **Barrel files (`index.ts`) do not benefit from it** — every build rewrites them, even when the input document has not changed and the barrel's bytes are byte-for-byte identical.

The result is that a rebuild always touches one file per barrel directory plus the root barrel, no matter how small the change. Anything watching the output re-runs over them.

This is separate from #3859 (which was about the formatter pass) and reproduces with `output.format: false`.

## Reproduction

Minimal setup — two schemas, two operations, `pluginOas` + `pluginTs` only, no formatter:

```js
import { build } from '@kubb/core';

const config = {
  root: process.cwd(),
  input: { path: './spec.json' },
  output: { path: './out', clean: false, format: false },
  plugins: [pluginOas({ validate: false }), pluginTs({ output: { path: 'types' } })],
};

await build({ config });          // first build
const first = stampMtimesAndHashes();
await build({ config });          // identical second build
const second = stampMtimesAndHashes();
```

Result:

```
files: 8
rewritten on an identical second build: 2
  of those, content actually differed: 0
    index.ts
    types/index.ts
```

The six generated modules are correctly skipped. Both barrels are rewritten with identical content:

```ts
export type { ListOwners200, ListOwnersQuery, ListOwnersQueryResponse } from "./ListOwners.ts";
export type { ListPets200, ListPetsQuery, ListPetsQueryResponse } from "./ListPets.ts";
export type { Owner } from "./Owner.ts";
export type { Pet } from "./Pet.ts";
```

## Impact

We generate an API client into an app that a dev server watches, and regenerate it when the OpenAPI document changes. Measured on a real 178-file client, changing **one property on one DTO**:

| | files |
| --- | --- |
| touched | 14 |
| content actually changed | **2** (the DTO's type and its faker factory) |

The remaining 12 are barrels and other always-rewritten files. Because the writes land over roughly a second, the bundler's watcher splits them into several separate compilations, so a one-property change costs four rebuild cycles instead of one.

The barrels are the part reproduced above in isolation; the rest of that set is under investigation on our side.

## Cause, as far as I can tell

Generated modules go through `write()`, which reads the file back and returns early when the content matches (`dist/index.js:420-434`). The barrel is assembled separately — `BarrelManager.getFiles()` builds a fresh file object per build, and the root barrel is constructed at `dist/index.js:1711-1740` — so whatever writes it does not appear to take the same path, or is handed content that differs before normalisation.

I have not tried to pin down which of those it is; the reproduction above should make it quick for someone who knows the write pipeline.

## Suggested fix

Route the barrel write through the same unchanged-content check the generated modules already use, so an unchanged barrel is skipped like everything else.

Happy to open a PR if you can point me at the intended place for it.

stijnvanhulle:
## 🎯 Changes

Fixes two related bugs where `write()`'s unchanged-content check fails to skip a file whose content hasn't actually changed, causing spurious rewrites and file-watcher churn on every rebuild.

### #3867 — Barrel files are rewritten on every build even when their content is unchanged

**Root cause found in this codebase:** `packages/core/src/build.ts`'s `setup()` called `fabric.use(fsPlugin)` (from `@kubb/react-fabric/plugins`). `fsPlugin.install()` registers its **own** raw `file:processing:update` listener that writes straight to disk with its own unchanged-content check — entirely separate from (and running *before*) the `storage.setItem` listener `build.ts` itself registers, which is the one that goes through `@kubb/core`'s deduped `write()` (`internals/utils/src/fs.ts`).

fsPlugin's own check compares the **untrimmed** freshly generated text against the **trimmed** content already on disk:
```js
if ((await readFile(resolve(path), 'utf-8'))?.toString() === data?.toString()) return
```
For a normal file this happens to match (the generated text has no surrounding whitespace). For a **barrel file** — `sources: []` — the typescript parser's `[banner, importsExportsPrint, source, footer].filter(s => s != null).join('\n')` includes `source` as an empty *string* (not `null`/`undefined`, so it survives the filter), which appends an extra trailing `\n` that's absent from what's on disk (already trimmed by a previous write). That mismatch is permanent, so fsPlugin unconditionally rewrites every barrel file on every build — confirmed with a build-twice repro in `packages/core/src/build.test.ts` before the fix (mtime changes every time), and it also silently bypasses `output.storage` (a custom storage or `output.write: false` would never see the barrel at all).

**Fix:** `build.ts` no longer installs `fsPlugin`. It drives `fabric.context.fileManager.write()` directly (the method `fsPlugin`'s `inject()` merely wrapped), so every file — barrel or not — goes through the single, correctly-deduped `storage.setItem` → `write()` path.

### #3859 — `output.format` runs after writing, so `write()`'s unchanged-content check never matches

**Root cause found in this codebase:** on this branch, `output.format` is handled entirely in `packages/cli/src/runners/generate.ts` (`@kubb/core`'s `build()`/`safeBuild()` never reference `output.format`), which runs the formatter as an external CLI pass over the whole output directory *after* `safeBuild()` has already written every file. A formatter's reflow (e.g. Kubb's printer omits the trailing semicolon Prettier/oxfmt add back, or reorders imports) permanently changes bytes on disk that the next build's freshly generated (pre-format) content can never match again — every file is rewritten, and reformatted, on every build once a formatter is configured. Confirmed with `oxfmt`/`prettier` directly on Kubb-shaped output.

**Fix (the issue's own lighter-weight "manifest" option):** `generate.ts` wraps `output.storage` (when unset and `output.format` is configured) with `formatCacheStorage` (`packages/cli/src/utils/formatCacheStorage.ts`), which keeps a manifest of the pre-format content hash last written per path, persisted at `<output>/.kubb/format-cache.json`. When the freshly generated content hashes the same as last time, the underlying write is skipped entirely — leaving the previously-formatted file untouched — instead of rewriting it with pre-format bytes that the formatter then has to reflow again. The manifest survives across separate `kubb generate` invocations, not just within one watch-mode process.

Verified end-to-end against the real `examples/typescript` fixture with real `pluginOas`/`pluginTs` and real `prettier`: a second `kubb generate` with no input change now touches **zero** files (mtimes identical, including the manifest); a real schema change touches only the affected model file.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (`@kubb/core` and `@kubb/cli` patch).

Fixes #3867
Fixes #3859

---
_Generated by [Claude Code](https://claude.ai/code/session_0114exixPFjYqtWT4gDNNLLH)_


stijnvanhulle:
The `Prerelease` check failure is unrelated to this PR's changes — the `Publish preview packages` step failed on a `pkg.pr.new` server error (500, then a 404 "There is no workflow defined for Cq44DlPqgm"), which is a transient issue on that external preview-publishing service, not a build/test/lint/typecheck failure. All other checks (Build, Tests, Typecheck, Linting, Spell Check, Performance benchmarks) passed. I don't have permission to re-run the job from here (`403 Resource not accessible by integration`) — a maintainer will need to re-trigger it, or it may resolve on its own on the next push.

---

PR #3925, opened 2026-08-26
feat(devtools): proof of concept devtools built on devframe

stijnvanhulle:
## 🎯 Changes

A proof of concept answering *can we add devtools to Kubb, and is it worth building properly?*

Adds a private `@kubb/devtools` package that surfaces a run in the browser — the parsed AST, the per-plugin pipeline, and the generated files. Run it with `KUBB_DEVTOOLS=1 kubb generate --watch`.

**No new instrumentation was needed in the build pipeline.** `Hookable<KubbHooks>` already emits everything the tool reads, and the CLI runner already owns the emitter, so the collector is just a new subscriber. `collector.ts` takes a `Hookable` and nothing else, which is what would make a Vite host cheap to add later (`unplugin-kubb` already builds its own emitter).

### The one real design constraint

`KubbDriver.dispose()` nulls `inputNode` at the end of every build, so nothing can read the canonical AST afterwards. And `kubb:generate:schema` fires once per *(node × plugin)* with that plugin's `override` already applied, so it isn't a substitute — it's a per-plugin view, not the tree.

So there are two sources: the host snapshots `driver.inputNode` at `kubb:build:start`, and the per-plugin views come from the walk. Diffing them is the feature — selecting a plugin shows which schemas it received against which its `include`/`exclude` filtered out. That's the "why didn't this schema generate" question you can't answer today.

### Scope

Deliberately not shippable: private package, no changeset, no exports map, no core API changes, and an env var rather than a `--devtools` flag. Deferred items (publishable package, a `KubbBuildStartContext` addition that would make the collector fully hooks-only, the Vite host, agent/MCP) are noted but not built.

`@kubb/devtools` is a devDependency of `@kubb/cli` and imported dynamically behind the env gate, so it stays out of the published dependency graph entirely.

### Styling

One file copied — `tokens.css` from `kubb-labs/platform`, with a provenance header; that repo stays the source of truth. `@kubb/ui` is `private: true` and ships uncompiled source, so a cross-repo dependency isn't possible, and the SPA is prebuilt into `clientAssets` so the copy is build-time only.

No component library and no utility framework: every panel is styled from those tokens directly. An earlier revision wired up Nuxt UI to match Studio, but the SPA turned out to reference no `U*` component and no Tailwind class, so it was removed — the screenshots are unchanged without it, and the workspace drops from 929 to 720 packages.

### Two notes for reviewers

- **Unrelated finding:** `KubbDriver.setup()` builds a fresh `NormalizedPlugin` and never copies `rawPlugin.generators`, so a statically-declared `generators: [...]` is silently dropped — the plugin is skipped with no error, and only `ctx.addGenerator()` works. Since `Plugin.generators` is a documented optional field, either the type is misleading or normalization has a bug. Not fixed here; it's a core behavior change.
- **Dependency wiring:** `vite` is pinned to `^8.2.2`, because a `^7` devDependency broke `unplugin-kubb`'s typecheck through vite-version cross-contamination. `vue-demi` is added to `allowBuilds`.

Verified end to end against `packages/adapter-oas/mocks/petStore.yaml`: server mounts and survives watch rebuilds, 13 schemas / 21 operations captured, all three panels populate, and a plugin scoped to `Pet` correctly shows 3 received against 10 skipped.

## ✅ Checklist

`pnpm test` (1484 tests, 93 files), `pnpm typecheck` (13/13), `pnpm lint`, and `oxfmt` are all clean.

## 🚀 Release Impact

Neither applies. `@kubb/devtools` is `private: true` and never published. The only published package touched is `@kubb/cli`, which gains a dormant, env-gated code path and a devDependency — no API or behavior change for anyone not setting `KUBB_DEVTOOLS=1`.

Worth knowing before merge: `.changeset/config.json` has `fixed: [["@kubb/*", "kubb", "unplugin-kubb"]]`, so a patch changeset here would version-bump and release **every** `@kubb` package for a proof of concept. Left out deliberately; it's a one-file addition if you'd rather have the bump.


stijnvanhulle:
Pushed 4f46f51 for a release-blocking bug in my own diff, and notes on the two bot findings.

### Fixed: `@kubb/devtools` was in the published dependency graph

`@kubb/cli` is published and public, but declared a **runtime** dependency on `@kubb/devtools`, which is `private: true` and never published. On release pnpm rewrites `workspace:*` to the real version, so `npm i @kubb/cli` would have 404'd on `@kubb/devtools@0.0.0`.

Moved to `devDependencies`, with the dynamic import routed through a helper that reports failure as a warning and lets the build continue — a published install has no `@kubb/devtools`, and `KUBB_DEVTOOLS=1` there should say so rather than abort a generation run. Verified the published manifest no longer names it.

### Dependency Diff: 231 new deps / +107.4 MB

Real, and worth a decision. Effectively all of it is `@nuxt/ui`, which drags in tiptap, prosemirror, yjs, framer-motion, reka-ui and embla-carousel.

Two things bound the impact:

- They are **devDependencies of a private, never-published package**. Nothing reaches a consumer of any `@kubb/*` package — the SPA is prebuilt into `clientAssets`, so the published surface is unaffected. The cost is contributor install time and disk.
- `devframe` itself is only 833 kB of that total.

But the honest reading: **the SPA currently uses zero Nuxt UI components** (no `U*` tag appears anywhere in `client/src`). It's registered via `@nuxt/ui/vite` and its CSS layer is imported, but every panel is hand-built on the copied `tokens.css`, and the screenshots confirm those already match Studio. So right now Nuxt UI is close to pure cost. Dropping it would remove most of the 231 deps and lose nothing visible; keeping it makes sense only if the panels are expected to grow into real Nuxt UI components. Flagging it for the maintainer rather than unilaterally reverting a deliberate choice.

### Changeset

Deliberately none. `@kubb/devtools` is private, and after the fix above the only published change is a dormant, env-gated code path in `@kubb/cli`.

Worth noting for whoever merges: `.changeset/config.json` has `fixed: [["@kubb/*", "kubb", "unplugin-kubb"]]`, so a patch changeset here would version-bump and release **every** `@kubb` package for a proof of concept. If you'd rather have the bump anyway, it's a one-file addition — your call, since it's your release process.

All 13 checks were green on 2486218; the push above is validated locally with `pnpm test` (1484 tests), `pnpm typecheck` (13/13), `pnpm lint`, and `oxfmt`.

---

stijnvanhulle:
Dropped Nuxt UI in 1a18d30, per the finding above.

Checking before removing it: the SPA referenced **no `U*` component and no Tailwind utility class** — every panel was already styled from the copied `tokens.css`. So `tailwindcss` went too (it was only pulling in the lightningcss and oxide binaries), and `vue-router` with it, since that was only present because Nuxt UI's optional peer stub fails to resolve under rolldown.

| | before | after |
| --- | --- | --- |
| workspace packages | 929 | 720 |
| client CSS | 214 kB | 17 kB |
| client JS | 178 kB | 133 kB |

The screenshots are unchanged — same brand orange, tokens, monospace and selection states — which is the confirmation that the library was contributing nothing here. If the panels later want real Nuxt UI components, adding it back is a one-line plugin registration.

Re-verified end to end after the change: server mounts, all three panels populate, AST tree renders, file contents read from disk, and `plugin-pet-only` still shows 3 received against 10 skipped. `pnpm test` (1484), `pnpm typecheck` (13/13), `pnpm lint` and `oxfmt` clean.

PR description updated to match.

---

stijnvanhulle:
Pushed 9490821 for the new dependency-diff finding (+67 deps / +28.8 MB from shiki).

Traced it precisely rather than guessing: `@shikijs/langs` (8.7 MB) and `@shikijs/themes` (1.5 MB) are each a **single package bundling every language and theme shiki knows** — using one grammar costs the same as using all 150+, so that part is a genuine floor for using shiki's per-language API at all, not something this PR can trim further.

What *was* avoidable: the umbrella `shiki` package I depended on unconditionally requires `@shikijs/engine-oniguruma` (a wasm grammar scanner), even though `highlight.ts` only ever imports the JS engine. Switched to depending directly on the four scoped packages this code actually uses — `@shikijs/core`, `@shikijs/engine-javascript`, `@shikijs/langs`, `@shikijs/themes` — which drops `@shikijs/engine-oniguruma` and the `shiki` wrapper package itself, since neither of those four requires it.

One thing worth recording so it isn't mistaken for the same kind of dead weight as the earlier Nuxt UI removal: `@shikijs/engine-javascript` itself still pulls in `oniguruma-to-es` (plus its own `oniguruma-parser`/`regex`/`regex-recursion` chain). That part is **not avoidable** — it's what lets the JS engine interpret oniguruma-flavored TextMate grammar syntax without a wasm binary at all; every TextMate-based highlighter needs one of the two (wasm-native or this transpiler chain).

Browser output is byte-identical (same chunk hashes, confirmed) — this only changes what `pnpm install` puts on disk. Re-verified end to end after the swap: highlighting, pipeline badges, and file content all render the same as before.

---

stijnvanhulle:
Confirmed on 9490821: 67→65 new deps, 28.8 MB→27.6 MB — the drop matches removing `shiki` (602.9 kB) + `@shikijs/engine-oniguruma` (643.9 kB) almost exactly.

Checked whether the remaining `hast-util-to-html`/`mdast`/`micromark`/`unist` tree (~700–900 kB) is trimmable the same way: it isn't. It's a **direct** dependency of `@shikijs/core` itself (`pnpm-lock.yaml:4918`), not something pulled in by the `codeToHtml` call specifically — `@shikijs/core` is monolithic the same way `@shikijs/langs`/`@shikijs/themes` are, so it installs in full regardless of which of its exports get used. Switching to `codeToTokensBase` and hand-rolling HTML serialization to dodge it wouldn't remove a single byte from `node_modules`.

So 65 deps / 27.6 MB is the real floor for using `@shikijs/core` + per-language grammars at all in this codebase, not a gap I left on the table. Not chasing this further — the alternative would be hand-writing a tokenizer to avoid shiki entirely, which is disproportionate for this POC and would also break from the syntax highlighter `kubb-labs/platform` already standardizes on for the same purpose.

---

stijnvanhulle:
Opened #3926 to write the idea up on its own, with screenshots, separate from this PR's implementation log. This PR is the tryout referenced there.

---

PR #3935, opened 2026-08-27
fix(plugin-barrel): report duplicate barrel exports instead of emitting an unparseable file

xeoneux:
A barrel that re-exports the same name from two modules is a duplicate binding, which is a parse
error rather than a type error. The file was emitted regardless, so the failure surfaced from
whatever eventually read the barrel — pointing at the generated file, not at the two sources.

`makeBarrel` now filters repeated names and reports each collision:

```
[KUBB_BARREL_DUPLICATE_EXPORT]: "createCancelJourneyInstanceResponse" is exported by both
"./createCancelJourneyInstance.ts" and "./createCancelJourneyInstanceResponse.ts", so
".../generated/journey-execution/faker/index.ts" cannot re-export both.
```

The check lives in `makeBarrel` because every barrel — per plugin, nested, recursive, and root —
is constructed there, so no path can bypass it.

Adds the `KUBB_BARREL_DUPLICATE_EXPORT` diagnostic code with a catalog entry.

### Behaviour

- The name is emitted **once**, from the first module that declared it, so the barrel parses.
- Each collision is reported at `error` severity with both file paths.
- Wildcard exports pass through untouched, since `export * from './x'` carries no names to compare.

Added a case to `plugin-barrel/src/utils.test.ts` with two files exporting the same name, asserting
both that the name appears exactly once in the emitted barrel and that the diagnostic names both
files.


xeoneux:
Fixes #3934

packages/core/src/constants.ts
```diff
@@ -132,6 +132,11 @@ export const diagnosticCode = {
    * The linter pass over the generated files failed.
    */
   lintFailed: 'KUBB_LINT_FAILED',
+  /**
```
stijnvanhulle, line 135: This will also need a new page in https://github.com/kubb-labs/docs/tree/main/docs/5.x/reference/diagnostics and the main/overview page: https://github.com/kubb-labs/docs/blob/main/docs/5.x/reference/diagnostics.md

packages/plugin-barrel/src/utils.ts
```diff
@@ -105,11 +106,60 @@ function isBarrelPath(path: string): boolean {
...
+
+  return deduplicated.filter((node) => node.name === undefined || node.name === null || !Array.isArray(node.name) || node.name.length > 0)
+}
+
 function makeBarrel(dirPath: string, exports: Array<ExportNode>): FileNode {
   return ast.factory.createFile({
     baseName: 'index.ts',
     path: `${dirPath}${BARREL_SUFFIX}`,
-    exports,
+    exports: withoutDuplicateExports(dirPath, exports),
```
stijnvanhulle: Could we do this before the `makeBarrel` so we dont need all those extra loops to validate?

xeoneux: Refactored barrel export collection into a single pass:
- Pure `makeBarrel`: Made `makeBarrel` a pure constructor by removing post-creation validation and filter loops.
- Single-pass collector: Deduplication and collision reporting now happen directly while walking leaf files for a directory, creating `ExportNode`s with only the unique names and eliminating intermediate AST allocations.
- Zero overhead on wildcards: `barrelType: 'all'` skips duplicate tracking and name map allocations entirely.

packages/core/src/constants.ts
xeoneux: https://github.com/kubb-labs/docs/pull/191

packages/plugin-barrel/src/utils.ts
```diff
@@ -120,13 +121,31 @@ function makeBarrel(dirPath: string, exports: Array<ExportNode>): FileNode {
   })
 }
 
-type LeafContext = {
-  dirPath: string
-  leafPath: string
-  sourceFile: FileNode | null
+type NamedScope = {
+  valueOwners: Map<string, string>
+  typeOwners: Map<string, string>
```
stijnvanhulle, line 126: Splitting values from types lets through the case the split was meant to allow. TypeScript puts a re-exported type and a re-exported value in the same slot:

```ts
// index.ts
export { Pet } from './a.ts'
export type { Pet } from './b.ts'
```

```
index.ts(1,10): error TS2300: Duplicate identifier 'Pet'.
index.ts(2,15): error TS2300: Duplicate identifier 'Pet'.
```

Same two errors when both come from one module. So `Pet` as a const in `a.ts` and `Pet` as a type in `b.ts` still produces a barrel that fails, and this code reports nothing.

One `Map<string, string>` keyed by name is correct here, and it drops half of `collectNamedLeafExports`.

stijnvanhulle (commented):
Right instinct and the right stage of the pipeline to catch this. Three things block it, and the implementation is bigger than the description promises.

I applied the branch locally and checked each claim against `tsc` and a scratch test. All 53 plugin-barrel and Diagnostics tests pass, the changeset is accurate, and files still write at `KubbDriver.ts:397` after `kubb:plugins:end`, so the deduped barrel does land on disk as you describe.

Details inline.

packages/plugin-barrel/src/utils.test.ts
```diff
@@ -50,6 +51,97 @@ describe('getBarrelFiles', () => {
...
+    const exported = barrels.flatMap((barrel) => barrel.exports.flatMap((node) => (Array.isArray(node.name) ? node.name : [])))
+    expect(exported.filter((name) => name === 'createPetResponse')).toHaveLength(1)
+
+    const collision = diagnostics.find((diagnostic) => diagnostic.code === Diagnostics.code.barrelDuplicateExport)
+    expect(collision?.message).toContain('createPetResponse')
+    expect(collision?.message).toContain('createPet.ts')
+    expect(collision?.message).toContain('createPetResponse.ts')
+  })
+
+  it('allows a value export and a type export with the same name without collision', () => {
```
stijnvanhulle, line 74: This test asserts the bug from the two-map split, so it will keep the fix out. `export { Pet } from './pet.ts'` next to `export type { Pet } from './pet.ts'` is `TS2300: Duplicate identifier 'Pet'` on both lines. Worth flipping to expect one collision diagnostic and one surviving export.

packages/plugin-barrel/src/utils.ts
```diff
@@ -207,7 +243,16 @@ function* walkAllOrNamed(node: BuildTree, params: LeafWalkParams, isRoot: boolea
 
   if (!isRoot && !params.recursive) return subtreeLeaves
 
-  const exports = subtreeLeaves.flatMap((leafPath) => params.strategy({ dirPath: node.path, leafPath, sourceFile: params.sourceFiles.get(leafPath) ?? null }))
+  let exports: Array<ExportNode>
+  if (params.barrelType === 'named') {
+    const scope = createNamedScope()
+    exports = subtreeLeaves.flatMap((leafPath) => collectNamedLeafExports(node.path, leafPath, params.sourceFiles.get(leafPath) ?? null, scope))
+    if (scope.collisions.length > 0) {
+      reportCollisions(node.path, scope.collisions)
```
stijnvanhulle, line 251: Each barrel gets a fresh scope, so one collision is reported once per barrel that spans both files. Plugin barrels run with `recursive: true` (`plugin.ts:167`) and the root barrel spans the whole output tree, so the minimum is two.

I ran it on `out/faker/{a,b}.ts` both exporting `dup`:

```
plugin barrels: [ '/out/faker/index.ts' ]  root barrels: [ '/out/index.ts' ]
total diagnostics: 2
```

That is your own example from the PR body, printed twice. In a deeper tree it is once per ancestor barrel. A seen-set of `name + first + second` across the run, or reporting only where the collision first appears, would keep it to one.

packages/plugin-barrel/src/utils.ts
```diff
@@ -148,12 +167,12 @@ function partitionIndexableNames(sources: ReadonlyArray<SourceNode>): Map<boolea
   return byTypeOnly
 }
 
-const allStrategy: LeafStrategy = ({ dirPath, leafPath, sourceFile }) => {
+function collectAllLeafExports(dirPath: string, leafPath: string, sourceFile: FileNode | null): Array<ExportNode> {
```
stijnvanhulle, line 170: `barrelType: 'all'` has the same collision and walks past it. Two star exports sharing a name is an error too, just a different one:

```
index.ts(2,1): error TS2308: Module './a.ts' has already exported a member named 'dup'. Consider explicitly re-exporting to resolve the ambiguity.
```

The name also silently disappears from the barrel's surface at runtime, which is nastier to debug than a parse error.

The PR body says wildcards carry no names to compare, but `index.sourceFiles` has them and this function already reads `sourceFile.sources` for `hasOnlyNonIndexableSources`. You cannot drop a name from `export *`, but you can report the collision. Otherwise half the config surface stays broken.

packages/plugin-barrel/src/utils.ts
```diff
@@ -120,13 +121,31 @@ function makeBarrel(dirPath: string, exports: Array<ExportNode>): FileNode {
   })
 }
 
-type LeafContext = {
-  dirPath: string
-  leafPath: string
-  sourceFile: FileNode | null
+type NamedScope = {
```
stijnvanhulle, line 124: The PR body says the check lives in `makeBarrel` because every barrel is built there. It does not. It lives in both walkers, with `NamedScope` threaded through `collectNamedLeafExports`, and `LEAF_STRATEGIES` was deleted in favor of the same `barrelType === 'named'` branch written twice.

`makeBarrel(dirPath, exports)` already receives every `ExportNode` for the directory, each carrying `name` and `isTypeOnly`. A single loop there filtering names already seen does the whole job: no scope type, no `createNamedScope`, no `reportCollisions` helper, and the strategy map stays. It also makes the value/type mistake above hard to write, because there is only one set to check against.

Roughly 15 lines in one function instead of +60/-26 across two walkers, and it matches what the description already claims.

packages/plugin-barrel/src/utils.ts
```diff
@@ -167,24 +186,41 @@ const namedStrategy: LeafStrategy = ({ dirPath, leafPath, sourceFile }) => {
     return [ast.factory.createExport({ path: modulePath })]
   }
 
+  const keptValues: Array<string> = []
+  for (const name of valueNames) {
+    const owner = scope.valueOwners.get(name)
+    if (owner === undefined) {
+      scope.valueOwners.set(name, modulePath)
+      keptValues.push(name)
+    } else if (owner !== modulePath) {
```
stijnvanhulle, line 195: `owner !== modulePath` can never be false. `partitionIndexableNames` dedupes names within a file, and each leaf path appears once per barrel, so `owner` and `modulePath` never match. Same on line 206.

packages/plugin-barrel/src/utils.ts
```diff
@@ -120,13 +121,31 @@ function makeBarrel(dirPath: string, exports: Array<ExportNode>): FileNode {
...
+}
+
+function reportCollisions(dirPath: string, collisions: ReadonlyArray<{ name: string; first: string; second: string }>): void {
+  for (const { name, first, second } of collisions) {
+    Diagnostics.report({
+      code: Diagnostics.code.barrelDuplicateExport,
+      severity: 'error',
+      message: `"${name}" is exported by both "${first}" and "${second}", so "${dirPath}${BARREL_SUFFIX}" cannot re-export both.`,
+      help: 'Rename one of the colliding declarations, or resolve the collision through the plugin resolver that names them.',
+      location: { kind: 'config' },
```
stijnvanhulle, line 145: `kind: 'config'` points the reader at their Kubb config, but the collision is between two generated files and the config is usually innocent. `location` is optional, so omitting it reads truer than pointing at the wrong place.

stijnvanhulle:
@all-contributors please add @xeoneux for code

stijnvanhulle:
@xeoneux Can you have another look at the comments? Thanks

stijnvanhulle:
Superseded by #4043, which reapplies the fix on current main and addresses every review thread. The diagnostic documentation remains in kubb-labs/docs#191.

PR #3936, opened 2026-08-27
feature: First version of reused @kubb/studio package for connection between cli and kubb.studio

stijnvanhulle:
## ✅ Checklist

## 🚀 Release Impact


packages/cli/src/index.ts
```diff
@@ -31,18 +31,24 @@ export async function run(argv: Array<string> = process.argv): Promise<void> {
   const { command: generateCommand } = await import('./commands/generate.ts')
   const { command: validateCommand } = await import('./commands/validate.ts')
   const { command: mcpCommand } = await import('./commands/mcp.ts')
+  const { definition: studioDefinition } = await import('./commands/studio.ts')
+  // The runner pulls in @kubb/studio, an optional peer, so it loads only when `kubb studio` runs.
```
stijnvanhulle: Do the same for mcp and validate

packages/cli/src/runners/studio/credentials.ts
```diff
@@ -0,0 +1,71 @@
...
+import { homedir } from 'node:os'
+import path from 'node:path'
+import process from 'node:process'
+
+/**
+ * Root for everything the CLI persists between runs: the paired credential, the machine secret,
+ * the last Studio config, and the live-session registry. `KUBB_HOME` relocates all of it.
+ */
+export function getKubbHome(): string {
+  return process.env.KUBB_HOME ?? path.join(homedir(), '.kubb')
```
stijnvanhulle, line 12: Add this to env files here and platform repo

packages/cli/src/runners/studio/credentials.ts
```diff
@@ -0,0 +1,71 @@
+import { chmod, mkdir, readFile, rm, writeFile } from 'node:fs/promises'
```
stijnvanhulle: Could we use our fs helper in @internals instead?

packages/cli/src/runners/studio/run.ts
```diff
@@ -0,0 +1,386 @@
...
+   * Current `@kubb/cli` version, reported to Studio and used for the telemetry payload.
+   */
+  version: string
+  configPath?: string
+  /**
+   * Base URL of the Studio instance, for a self-hosted deployment. Resolved before it reaches here,
+   * since stored credentials are bound to it.
+   */
+  studioUrl: string
+  allowWrite: boolean
```
stijnvanhulle: allowWrite: boolean
  /**
   * Whether Studio may change plugin options in the project's `kubb.config.ts`.
   */
  allowConfigEdit: boolean
  allowInput: boolean
  allowExec: boolean

can those all become part of `permission` object in StudioOptions?

packages/cli/src/runners/studio/run.ts
```diff
@@ -0,0 +1,386 @@
...
+  /**
+   * Whether Studio may change plugin options in the project's `kubb.config.ts`.
+   */
+  allowConfigEdit: boolean
+  allowInput: boolean
+  allowExec: boolean
+  /**
+   * Whether to open the approval page in a browser during pairing.
+   */
+  open: boolean
```
stijnvanhulle: rename it to autoOpen (same on other places)

packages/cli/src/runners/studio/run.ts
```diff
@@ -0,0 +1,386 @@
...
+
+/**
+ * Connects this project to Studio and streams generation events until the process is stopped.
+ */
+async function connect(options: StudioOptions, retryPairing = true): Promise<void> {
+  const envToken = process.env.KUBB_AGENT_TOKEN
+  const stored = envToken ? null : await readCredentials()
+  // A credential is only reused for the Studio it was issued by, so switching `--url` re-pairs
+  // instead of sending one instance's token to another.
+  let credentials: Credentials | null = envToken
```
stijnvanhulle: instead of using let and 2 short if function, use IEFEE like const credentials = (() => {//do logic and return)())

stijnvanhulle: Take this appraoch and use on more places

packages/studio/src/protocol/index.ts
```diff
@@ -0,0 +1,490 @@
...
+   */
+  plugins?: Array<{
+    name: string
+    options?: object
+  }>
+  /**
+   * Package names of plugins Studio explicitly turned off. Takes priority over the disk
+   * config's `plugins` array, so a plugin listed here is dropped even when present on disk.
+   */
+  disabledPlugins?: Array<string>
```
stijnvanhulle: can we add this to plugins

plugins?: Array<{
    name: string
    options?: object
    disabled?: boolean // add like this?
  }>

packages/studio/tsdown.config.ts
```diff
@@ -0,0 +1,36 @@
...
+  protocol: 'src/protocol/index.ts',
+}
+
+const shared: Partial<UserConfig> = {
+  platform: 'node',
+  sourcemap: true,
+  shims: true,
+  exports: true,
+  deps: {
+    neverBundle: [/^@kubb\/(?!core\/package)/],
```
stijnvanhulle: is this never bundle still needed as can use version from @kubb/studio internal itself?

stijnvanhulle: No, it's not needed. Nothing in `@kubb/studio` imports `@kubb/core/package`, and every other package's `tsdown.config.ts` already uses the plain `/^@kubb\//` pattern. Simplified it in 3dc467c.

---

packages/cli/src/runners/studio/credentials.ts
stijnvanhulle: Yes. Switched `readCredentials`/`writeCredentials`/`clearCredentials` to `@internals/utils`'s `read`/`write`/`clean`. Kept the explicit `mkdir(..., { mode: 0o700 })` and `chmod(file, 0o600)`, since those tighten permissions for the bearer token and neither helper does that. In `dea3e6f`.

---

packages/cli/src/runners/studio/run.ts
stijnvanhulle: Done. `StudioOptions` now carries a `permission: Record<Permission, boolean>` field instead of the four flat booleans. `resolvePermissions` and the `runner` mapping updated to match, tests too. In `dea3e6f`.

---

packages/cli/src/runners/studio/run.ts
stijnvanhulle: Renamed `StudioOptions.open` (and the `login()` destructure) to `autoOpen`. The `--open` CLI flag name itself is unchanged, since that's user-facing; it now maps onto `autoOpen` internally. Checked the rest of the codebase and this was the only place using the old name. In `dea3e6f`.

---

packages/cli/src/runners/studio/run.ts
stijnvanhulle: Done — `connect()` now resolves `credentials` through an IIFE instead of a `let` reassigned by two guard `if`s. In `dea3e6f`.

I looked for other `let` + guard-`if` spots to apply the same pattern to (both in this file and the rest of the `@kubb/studio` change), but everything else that uses `let` is either a genuine loop accumulator or a mutable flag across a WebSocket lifecycle — converting those to an IIFE wouldn't remove any `let`, since the mutation itself is the point. Let me know if there's a specific spot you had in mind.

---

packages/studio/src/protocol/index.ts
stijnvanhulle: Done in this package: `JSONKubbConfig['plugins']` entries now take `disabled?: boolean` and `disabledPlugins` is gone. `mergePlugins` in `resolveConfig.ts` and its call site in `connectStudio.ts` updated to match, tests updated. In `dea3e6f`.

One thing to flag: `apps/agent` and `apps/kubb.studio` in `kubb-labs/platform` have their own, separately-implemented copy of this `JSONKubbConfig`/`mergePlugins` (not importing this package) and its frontend still sends the old `disabledPlugins` array. That's the existing Docker-agent protocol, not the new CLI path this PR adds, so I left it alone rather than changing a live, already-shipped surface as a side effect here — but it'll want the same change, coordinated with a frontend update, before the two protocols are meant to converge.

---

stijnvanhulle: Update: the Docker-agent protocol is now synced too. In `kubb-labs/platform` (`feature/kubb-studio-cli-connect`, commit `ac83188`), `apps/kubb.studio`'s frontend sends disabled plugins as `plugins` entries with `disabled: true` instead of a separate `disabledPlugins` array, matching this package's protocol.

---

packages/cli/src/runners/studio/credentials.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Root for everything the CLI persists between runs: the paired credential, the machine secret,
+ * the last Studio config, and the live-session registry. `KUBB_HOME` relocates all of it.
+ */
+export function getKubbHome(): string {
+  return process.env.KUBB_HOME ?? path.join(homedir(), '.kubb')
+}
+
+const credentialsPath = () => path.join(getKubbHome(), 'credentials.json')
```
stijnvanhulle: 'credentials.json' can be a const

packages/studio/src/protocol/index.ts
```diff
@@ -0,0 +1,486 @@
+/**
+ * WebSocket message types for the agent ↔ Studio protocol.
+ *
+ * - Studio → agent: `command` (generate, connect, save), `pong`, `disconnect`
+ * - Agent → Studio: `connected`, `data`, `ping`, `config-saved`
+ * - Either way: `kubb:error`
```
stijnvanhulle: can we use everywhere kubb:NAME for the events like kubb:error? (update also platform repo)

packages/studio/src/protocol/index.ts
```diff
@@ -0,0 +1,486 @@
...
+
+import type { Config } from '@kubb/core'
+
+/**
+ * JSON-serializable Kubb config exchanged over the WebSocket. A live `kubb/kit` config holds
+ * functions and class instances that cannot survive JSON, so both sides pass this flattened shape
+ * and rebuild the real config from it.
+ */
+export type JSONKubbConfig = {
+  /**
```
stijnvanhulle, line 22: add this as part of  name: string or  options?: object instead of generic text for plugins

packages/studio/src/configFile.test.ts
```diff
@@ -0,0 +1,789 @@
+import { readFileSync } from 'node:fs'
```
stijnvanhulle, line 1: avoid usign toMatchInlineSnapshot

packages/studio/src/pair.ts
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Polls until the user approves or denies, honoring the server's `slow_down` back-off. A poll that
+ * cannot reach Studio is warned about and retried, since the code stays valid either way.
+ *
+ * Studio's own endpoint is used rather than the auth layer's `/device/token`, because an approved
+ * Kubb pairing is worth an agent bearer token, not a user session.
+ *
+ * @throws when the code expires, the user denies it, or Studio returns an unexpected error.
+ */
+export async function pollForPairingToken({ studioUrl = agentDefaults.studioUrl, session }: PollOptions): Promise<PairingResult> {
```
stijnvanhulle, line 122: can't we reuse axios, ofetch or ky for this to reduce extra code in our code base for polling?

packages/studio/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export { createClient, type Client, type ClientOptions } from './client.ts'
```
stijnvanhulle, line 1: can we reduce amount of items we expose based on what is neede in the platform repo? (agent docker)

packages/cli/src/runners/studio/credentials.ts
stijnvanhulle: Done in #3962. Lifted into `CREDENTIALS_FILENAME` and added `getCredentialsPath()`, so `run.ts` names the file through that instead of rebuilding `path.join(getKubbHome(), 'credentials.json')` a second time.

---

packages/studio/src/index.ts
stijnvanhulle: Done in #3962. `apps/agent/server/plugins/studio.ts` in the platform repo is the only importer, and it takes `createClient`, `defaultStudioUrl`, `InvalidAgentTokenError`, `pollForPairingToken`, `setStorage`, `startPairing`, and the `Client` type.

So I dropped `ClientOptions`, `PairingResult`, and `PairingSession`, which nothing imports. `createFileStorage` stays: it isn't used by the agent, but `@kubb/cli`'s `run.ts` calls it to point storage at `~/.kubb/cache`.

---

packages/studio/src/protocol/index.ts
stijnvanhulle: Done in #3962, and in the platform repo in kubb-labs/platform#400.

Every message on the wire is now `kubb:`-prefixed: `kubb:command`, `kubb:connected`, `kubb:data`, `kubb:config-saved`, `kubb:ping`, `kubb:pong`, `kubb:disconnect`. `kubb:error` was already there, and the hook names in `KubbHooks` were too, so the whole protocol sits in one namespace.

Left alone deliberately: the agent `status` values in `shared/schemas/api.ts` and `server/api/agents/index.get.ts` also spell `'connected'`, but that is a database column describing an agent, not a message.

This breaks the wire, so the two PRs need to land together with a matching `@kubb/studio` release. Worth noting the platform repo links `@kubb/studio` to the local kubb checkout, so it picks the rename up only after `pnpm --filter @kubb/studio build` there.

---

packages/studio/src/configFile.test.ts
stijnvanhulle: Done in #3962. Converted 41 of them to explicit `toStrictEqual` / `toBe` literals, across `configFile.test.ts`, `connectStudio.test.ts`, `resolveConfig.test.ts`, and `@kubb/cli`'s `run.test.ts`.

12 remain, all in `configFile.test.ts`, and all asserting a whole generated config file or a multi-line span out of one. That is the case `.claude/rules/testing.md` reserves snapshots for, where a hand-maintained literal is impractical. Happy to convert those too if you'd rather have none at all.

---

packages/studio/src/pair.ts
stijnvanhulle: Adopted `ofetch` in #3962, though one layer down from the poll loop.

I looked at what's on npm for the loop itself, and nothing fits. `p-retry` and `async-retry` retry when the function *throws*, but the decision here comes from the response body (`authorization_pending` retry, `slow_down` retry slower, `access_denied` fatal), and `slow_down` moves the interval mid-flight while both libraries fix their schedule at call time. `p-wait-for` and `promise-poller` are "call until truthy" with no fatal path, and both have been unpublished for ~3 years. `openid-client` does implement RFC 8628 properly, but it wants a real OIDC issuer config and this endpoint returns an agent bearer token rather than an OAuth token response.

Where `ofetch` did pay off is `api.ts`, and it took out more than the request:

- `postJson` is now a one-line wrapper. `ofetch` encodes and parses the JSON, and `ignoreResponseError` is exactly the existing `allowErrorResponse` flag.
- The `HttpError` class is gone. `FetchError` already carries `statusCode`, which is what `rejectedWith` reads.
- `runRegistration`'s hand-written `for` loop over `[0, 2s, 4s, 8s]` is gone, replaced by `retry` + `retryDelay`. A network error counts as a 500 there, so connection failures still retry.

Net −135 lines across the PR.

One behavior change to flag: registration now retries a network failure or a 5xx, and no longer a 403. `ofetch`'s default `retryStatusCodes` excludes it, which seems right, since a 403 means Studio refused the machine token and asking again with the same one can't succeed. It also drops the per-attempt "retrying" warning; the final error still names the failure. Say the word if you'd rather keep either.

I also checked the other `fetch` call sites in the repo and left them: `Telemetry.ts` is fire-and-forget with the response never read, `adapter-oas/load/source.ts` needs the raw `Response` for its diagnostics, and the update check in `generate/run.ts` would save one line at the cost of a new dependency on `@kubb/cli`.

---

packages/studio/src/protocol/index.ts
stijnvanhulle: Left as `disabled?: boolean` for now, per your call when I asked.

For the record on what's there: the entry is already `{ name: string; options?: object; disabled?: boolean }`, so `disabled` is a field on the plugin object rather than free text. If you did mean folding the flag into `name` (a marker prefix, say), happy to do that in a follow-up.

---

stijnvanhulle:
Splitting this into a stack of 5 smaller PRs so it's easier to review by concept, each based on the previous:

1. #3968 — shared utilities cleanup and CLI lazy-loading
2. #3969 — studio protocol and WebSocket client core
3. #3970 — device-authorization pairing
4. #3971 — config resolution and AST-based config editing
5. #3972 — the `kubb studio` CLI command

The last PR in the stack reproduces this PR's head exactly (verified with a direct diff), aside
from one stale changeset dropped as part of the split (`olive-parrots-shake.md`, superseded by
`studio-directional-protocol.md` in #3969). Closing this one in favor of the stack.

The companion PR in kubb-labs/platform (kubb-labs/platform#390) is being split the same way.

https://claude.ai/code/session_018oTZxx4JE5oY1Wb6NC4ojP

---

Issue #3981, opened 2026-09-05
Handle periodic reauthentication in @kubb/studio and the Studio CLI


stijnvanhulle:
Related: [Platform #426: server expiry and Docker support](https://github.com/kubb-labs/platform/issues/426).

## What should change

When Studio rejects an expired agent token, `@kubb/studio` should tell the CLI or Docker adapter that the user needs to approve access again. The CLI should guide the user through approval and reconnect.

Today, startup failures reach the caller of `createClient().connect()`. Later failures behave differently: the reconnect code catches `InvalidAgentTokenError`, logs it, and stops retrying. The CLI can then sit disconnected, waiting for Ctrl+C without offering a way to recover.

The server will decide how many days a token lasts. This issue covers the client package and CLI behavior when that token expires.

## Plan

- [ ] Add an optional callback to `createClient()`, such as `onAuthRequired`, for token rejection during background reconnect. Stop all connections and retry timers for that client, then notify the caller once. Several connections rejecting the same token must not start several pairing flows. Keep startup errors working as before and avoid handling the same failure twice.
- [ ] Let the CLI and Docker adapter own pairing and credential storage. The package reports the rejected token, and the caller obtains a replacement and creates a new client. Keep retrying ordinary network failures. A normal 24-hour session expiry or a revoked session alone must not trigger pairing.
- [ ] Handle rejected tokens at startup and while the CLI is running. For an interactive login, explain what happened, reuse the browser approval flow, save the new token, and reconnect. Keep saved project permissions only if the new login has the same agent identity and Studio URL.
- [ ] In CI or a noninteractive connection, exit with a nonzero status and instructions to run `kubb studio login`. If the token came from `KUBB_AGENT_TOKEN`, tell the operator to replace it. Do not overwrite it or switch to saved credentials. Keep the existing behavior of an explicitly requested `login` command.
- [ ] Cancel pairing and reconnect work on shutdown, and remove signal handlers. Add cancellation to the existing pairing helpers if needed. Stop repeated recovery attempts if Studio immediately rejects a newly approved token.
- [ ] Add the package changeset(s) and document the callback for the Docker adapter.

Start in `packages/studio/src/`: `client.ts`, `connectStudio.ts`, `api.ts`, and `pair.ts`. The CLI changes belong in `packages/cli/src/runners/studio/run.ts` and `credentials.ts`.

## Server agreement and release order

[...]

PR #3982, opened 2026-09-05
feat(studio,cli): reauthenticate automatically when Studio rejects a live token

stijnvanhulle:
## 🎯 Changes

Implements the client and CLI half of periodic reauthentication (kubb-labs/kubb#3981). The server half, which enforces token expiry, is tracked separately in kubb-labs/platform#426.

`@kubb/studio`:
- `createClient` gains an optional `onAuthRequired` callback, called once a live pool's token is rejected during background reconnect, after the whole pool is already stopped. It fires exactly once per client even when several pool sessions reject the same token at once. It never fires for a token rejected at startup (`connect()` still throws for that, unchanged) or for an ordinary session expiry or revocation.
- `startPairing` and `pollForPairingToken` take an optional `signal`, so a host can cancel an in-flight pairing or poll. Aborting rejects with the new `PairingCanceledError`, exported from the package.

`@kubb/cli` (`kubb studio`):
- Reacts to `onAuthRequired`: explains what happened, reopens the browser approval flow, and reconnects with the new token. Saved project permissions carry forward only when the reauthenticated agent has the same identity and Studio URL.
- An operator-supplied `KUBB_AGENT_TOKEN` is never replaced automatically; the CLI tells the operator to update it.
- CI or a run without a TTY exits nonzero with instructions to run `kubb studio login`, without touching stored or environment credentials.
- A newly approved token rejected again immediately is treated as a hard failure instead of pairing forever (one automatic reauth per run, across both the startup and live-rejection paths).
- `connect()` now owns a single `AbortController` for its whole run, armed once and torn down in `finally`, so Ctrl+C cancels an in-flight pairing or poll and a reauth never leaves behind a duplicate `SIGINT`/`SIGTERM` listener.

## ✅ Checklist

## 🚀 Release Impact

https://claude.ai/code/session_01EDaxW9rK7cneiFvbvMSfaB

---
_Generated by [Claude Code](https://claude.ai/code/session_01EDaxW9rK7cneiFvbvMSfaB)_

### DIFF
--- a/packages/cli/src/runners/studio/run.ts
+++ b/packages/cli/src/runners/studio/run.ts
@@ -235,7 +235,7 @@ async function loadConfigs(options: StudioOptions): Promise<{ configPath: string
  * browser is available to re-pair with, or the token this run already re-paired for was rejected
  * again.
  */
-type RejectedTokenReason = 'envToken' | 'nonInteractive' | 'exhausted'
+type RejectedTokenReason = 'envToken' | 'nonInteractive' | 'reauthExhausted'
 
 /**
  * Explains a rejected token to the operator. Never touches stored or environment credentials
@@ -246,7 +246,7 @@ function explainRejectedToken(error: InvalidAgentTokenError, reason: RejectedTok
     return new Error(`${error.message} Pair again and update KUBB_AGENT_TOKEN.`)
   }
 
-  if (reason === 'exhausted') {
+  if (reason === 'reauthExhausted') {
     return new Error(`${error.message} Studio rejected the newly approved token too. Run \`kubb studio login\` and try again.`)
   }
 
@@ -401,7 +401,7 @@ export async function connect(options: StudioOptions): Promise<void> {
         await clearCredentials()
 
         if (state.hasReauthenticated) {
-          throw explainRejectedToken(error, 'exhausted')
+          throw explainRejectedToken(error, 'reauthExhausted')
         }
 
         if (isCIEnvironment() || !canUseTTY()) {
@@ -453,7 +453,7 @@ export async function connect(options: StudioOptions): Promise<void> {
       }
 
       if (state.hasReauthenticated) {
-        throw explainRejectedToken(error, 'exhausted')
+        throw explainRejectedToken(error, 'reauthExhausted')
       }
 
       if (isCIEnvironment() || !canUseTTY()) {
--- a/packages/cli/src/runners/studio/run.ts
+++ b/packages/cli/src/runners/studio/run.ts
@@ -7,7 +7,16 @@ import { KUBB_CONFIG_FILENAME } from '@internals/shared'
 import { toError } from '@internals/utils'
 import type { CLIOptions, Config } from '@kubb/core'
 import { cliReporter, logLevel as logLevelMap } from '@kubb/core'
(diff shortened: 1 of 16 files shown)


packages/cli/src/runners/studio/run.ts
```diff
@@ -200,131 +231,278 @@ async function loadConfigs(options: StudioOptions): Promise<{ configPath: string
...
-    console.log(styleText('yellow', `${error.message} Pairing again...`))
+    // A rejection right after a fresh login means the newly approved token was no good either, so
+    // one automatic re-pair is all this ever attempts, whether the rejection lands at startup or
+    // once the session is already live. A second one is treated as a hard failure instead of
+    // pairing forever.
+    let hasReauthenticated = false
 
-    return connect(options, false)
-  }
+    for (;;) {
```
stijnvanhulle: Do not use a syntax like this

packages/cli/src/runners/studio/run.ts
```diff
@@ -200,131 +231,278 @@ async function loadConfigs(options: StudioOptions): Promise<{ configPath: string
...
+    let usingEnvToken = !!envToken
 
-    await clearCredentials()
+    const { allowWrite, allowConfigEdit, allowInput, allowExec } = await resolvePermissions(options, credentials, configPath, !usingEnvToken)
+    const granted = { allowWrite, allowConfigEdit, allowInput, allowExec }
 
-    if (isCIEnvironment() || !retryPairing) {
-      throw new Error(`${error.message} Run \`kubb studio login\` to pair again.`)
+    const detail = (label: string, value: string) => `${styleText('dim', label.padEnd(7))}  ${value}`
+    let hinted = false
```
stijnvanhulle: Everything that is state should be part of const state, so state.hinted

packages/cli/src/runners/studio/run.ts
```diff
@@ -200,131 +231,278 @@ async function loadConfigs(options: StudioOptions): Promise<{ configPath: string
...
-      hooks.hook('studio:connected', () => {
-        if (hinted || options.logLevel === 'silent') {
-          return
-        }
-        hinted = true
+  try {
+    const envToken = process.env.KUBB_AGENT_TOKEN
+    const stored = envToken ? null : await readCredentials()
+
+    let credentials: Credentials = await (async () => {
```
stijnvanhulle: Do not use let, why does it need let?

stijnvanhulle: Fixed in 7500172: `credentials`, `usingEnvToken`, `hinted`, and `hasReauthenticated` are now properties on one `const state` object instead of separate `let`s, and mutated via `state.x = ...`.

---

packages/cli/src/runners/studio/run.ts
stijnvanhulle: Fixed in 7500172: `hinted` is now `state.hinted`, alongside `state.credentials`, `state.usingEnvToken`, and `state.hasReauthenticated`.

---

packages/cli/src/runners/studio/run.ts
stijnvanhulle: Fixed in 7500172: swapped `for (;;)` for `while (true)`.

---

PR #4003, opened 2026-09-09
fix(kit): correct pnpm --no-git-checks flag in release scripts

stijnvanhulle:
## 🎯 Changes

`packages/kit/package.json`'s `release` and `release:stage` scripts passed `--no-git-check` to pnpm, which pnpm rejects since the flag is `--no-git-checks` (with the trailing "s"). Running either script failed with `unexpected argument '--no-git-check' found`. Both scripts now pass the correct flag.

## 🧪 How to test

- Step 1: `cd packages/kit`
- Step 2: Run `pnpm release:stage --dry-run` (or inspect the script with `pnpm run release:stage -- --help`).
- Step 3: Confirm pnpm no longer rejects the command with an unexpected-argument error.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change is for the docs (no release). This only touches the package's own release tooling scripts, not its published runtime code.
- [ ] This change is breaking, and the body says what breaks and what a user has to do.

https://claude.ai/code/session_01DPmiPhzLsaAMaQpBnt3vP6

---
_Generated by [Claude Code](https://claude.ai/code/session_01DPmiPhzLsaAMaQpBnt3vP6)_

### DIFF
--- a/configs/vitest.bench.config.ts
+++ b/configs/vitest.bench.config.ts
@@ -7,8 +7,5 @@ export default defineConfig({
   test: {
     include: ['**/*.bench.ts'],
     exclude: ['**/node_modules/**', '**/dist/**', '**/mocks/**'],
-    benchmark: {
-      include: ['**/*.bench.ts'],
-    },
   },
 })
--- a/packages/adapter-oas/src/adapter.bench.ts
+++ b/packages/adapter-oas/src/adapter.bench.ts
@@ -1,6 +1,6 @@
 import path from 'node:path'
 import { findCircularSchemasFromGraph } from '@kubb/ast'
-import { bench, describe } from 'vitest'
+import { test } from 'vitest'
 import { DEFAULT_PARSER_OPTIONS } from './constants.ts'
 import { parseDocument } from './load/normalize.ts'
 import { getSchemas } from './model/components.ts'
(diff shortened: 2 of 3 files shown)


stijnvanhulle:
The `Performance benchmarks (Vitest)` check fails on `14c8125` with `TypeError: bench is not a function` in `packages/adapter-oas/src/adapter.bench.ts:43`. This isn't caused by this PR's diff (it only touches `packages/kit/package.json`'s release scripts).

`main` is red the same way: the last three PR-workflow runs on `main` (`ci(changesets): version packages` twice, and `Update AGENTS.md`) all failed this same check before this PR existed. The pinned `vitest` catalog version (`^5.0.0`, resolved to `5.0.0` in the lockfile) no longer exports `bench` from `vitest` the way `adapter.bench.ts` imports it, so every PR against `main` currently fails this check regardless of its diff.

No fix exists yet for this on `main` or in an open PR. Fixing the benchmark file's vitest 5 API usage is outside this PR's scope (a one-line pnpm flag fix), so I'm leaving it as is rather than widening the diff. This isn't a flake, so re-running would not change the result.

---

stijnvanhulle:
The `Typecheck` check also fails on `14c8125`, with the same root cause as the benchmark failure above: `src/adapter.bench.ts(3,10): error TS2724: '"vitest"' has no exported member named 'bench'`. Not caused by this PR's diff either.

---

stijnvanhulle:
Pushed `fe850cb`, which fixes the root cause: vitest 5 dropped the top-level `bench()`/`describe()` benchmark API in favor of a `bench` function on the test context. `packages/adapter-oas/src/adapter.bench.ts` now uses `test('...', ({ bench }) => bench(name, fn).run(runOptions))`, and `configs/vitest.bench.config.ts` drops the `benchmark.include` key, which vitest 5 no longer recognizes.

Verified locally: `pnpm typecheck` and `pnpm exec vitest bench --config ./configs/vitest.bench.config.ts --run` both pass in `packages/adapter-oas`, and the package's regular test suite (533 tests) is unaffected.

---

PR #4028, opened 2026-09-14
feat(studio): add a studio:ready handshake acknowledgement

stijnvanhulle:
## 🎯 Changes

Adds a `studio:ready` acknowledgement so a host can tell a socket being open apart from Studio actually registering the connection for job dispatch. `StudioSession` waits up to 10 seconds for it after sending `agent:connect`, fires a new `studio:ready` hook once it lands (warning instead of failing against an older Studio that never sends one), and both CLI loggers print `✓ Ready to receive jobs`.

Part of the fix for kubb-labs/platform#535. The server side (sending the ack, gating dispatch on it) lands in a separate `kubb-labs/platform` PR that depends on this one being published first.

## 🧪 How to test

- Step 1: `pnpm --filter @kubb/studio test` and `pnpm --filter @kubb/cli test`
- Step 2: `pnpm typecheck && pnpm build` from the repo root
- Step 3: All packages typecheck, build, and their test suites pass

## ✅ Checklist

## 🚀 Release Impact

- [ ] This change is breaking, and the body says what breaks and what a user has to do.

## Summary by CodeRabbit

- **New Features**
  - Studio sessions now confirm readiness after connecting and display a “Ready to receive jobs” message in the CLI.
  - Added support for clients running directly within automated pipelines.
  - Exposed a session-ready event for integrations.

- **Bug Fixes**
  - Older Studio versions that do not send readiness confirmation now trigger a warning instead of failing.
  - Readiness waits are properly cleared once confirmation is received or the session is closed.


stijnvanhulle:
Fixed both CodeRabbit findings in 8ee9169:

- `#waitForReady` now returns early when `#disposed` is already true, so a session torn down while `loadConfig()` is still pending no longer arms a stray 10s timer.
- The timeout callback's `#warn` call is now wrapped with `Promise.resolve(...).catch(() => {})`, matching the existing `#emitError`/`#onOpen` pattern in this file, so a rejecting `studio:warn` listener can't produce an unhandled rejection.

Added a regression test for the first one: opens the socket, aborts the session while its config load is still pending, resolves the config, and asserts no timeout warning fires.

PR #4059, opened 2026-09-16
feat(studio)!: stop sending file contents on kubb:generation:end

stijnvanhulle:
## 🎯 Changes

`kubb:generation:end` now carries nothing. Reading generated source back was the one thing no `--allow-*` flag covered — a plain `kubb studio` session already streamed every file's full text to the browser on every run. This closes that gap, and as a side effect removes a memory cost nobody asked for: the agent no longer holds a whole run's output in RAM either.

### What moved where

| Data | Used to live on | Now lives on |
| --- | --- | --- |
| File list | `kubb:generation:end` (`storage`) | `kubb:build:end` (`files`), paths now relative to the agent's root |
| File count | `kubb:generation:end` (`storage` size) | `kubb:generation:summary` (`fileCount`) — always the accurate number, unlike the old count |
| File contents | `kubb:generation:end` (`storage`) | `studio:files` → `agent:files`, fetched per path, refused unless `allowRead` was granted |
| Peer dependencies | `kubb:generation:end`, round-tripped into `studio:snapshot` | `agent:snapshot` reply — the agent already resolves them while packing |
| `config` | `kubb:generation:end` | nowhere — a lossy `JSON.stringify` husk nothing read |

### Fetching a file, end to end

```
Agent (kubb studio)                         Studio UI
  │                                              │
  ├── kubb:build:end (paths) ──────────────────►│  tree fills
  ├── kubb:generation:end (empty) ──────────────►│  (nothing to read yet)
  │                                              │
  │                                              │◄─ user opens a file
  │◄───────────────── studio:files { paths } ────┤
  │
  ├─ allowRead not granted ─► agent:files { status: error }
  │
  ├─ allowRead granted:
  │    readItem(path) on the live Storage handle — only the requested paths
  │
  ├── agent:files { status: ok, files } ────────►│  editor shows the file
```

### Also in this PR

- **Memory.** The agent used to read every generated file into one `Record<string, string>` the moment a run finished, and kept that map alive for the whole session. It now keeps the live `Storage` handle and the run's path list instead, and reads a file's content back from `Storage` only when `studio:files` or `studio:snapshot` actually asks. A requested path is checked against that list before it reaches storage, so a caller still can't read anything outside what the run produced.
- **Snapshot.** `studio:snapshot` takes `bundledDependencies` in place of `peerDependencies`, so Studio's missing-dependency check still runs before any upload, sourced from the peer dependencies the agent resolves while packing.
- **`ClientInfo.kind`** drops `'ci'`. Its only reader was the now-deleted `skipStorage` flag, and a CI connection is really just the `kubb` binary, so it now reports `'cli'` like any other CLI run — which also fixes a latent bug where a refused CI input pointed at a CLI flag remedy instead of the environment variable that actually works for it.

## 🧪 How to test

- Step 1: `pnpm --filter @kubb/studio test && pnpm --filter @kubb/cli test`
- Step 2: `kubb studio` against a local Studio build on the matching protocol version, generate, and confirm the file tree fills with no source in the WebSocket frame
- Step 3: `kubb studio --allow-read`, open a file, confirm a `studio:files`/`agent:files` round trip fetches just that file
- Step 4: run a generation producing a few hundred files with `--allow-write`, watch agent process memory before and after this branch while never opening a file in the browser — it should no longer jump by the size of the generated output

## ✅ Checklist

## 🚀 Release Impact

- [x] This change is breaking, and the body says what breaks and what a user has to do.

https://claude.ai/code/session_01JWnTr5b7L8UeSQxFJPdUwC

---
_Generated by [Claude Code](https://claude.ai/code/session_01JWnTr5b7L8UeSQxFJPdUwC)_

## Summary by CodeRabbit

* **Breaking Changes**
  * Generation-complete events no longer include generated file contents or metadata.
  * Generated file paths are now reported relative to the project root.
  * Snapshot requests now specify bundled dependencies; dependency details are returned separately.

* **New Features**
  * Added read permission for retrieving generated file contents on demand.
  * Added `--allow-read` and `KUBB_AGENT_ALLOW_READ` options. Access is denied by default unless granted, with sandbox agents exempt.
  * File requests support up to 50 paths at a time.


packages/studio/src/protocol/index.ts
```diff
@@ -205,14 +205,12 @@ export type KubbHooks = {
...
-      peerDependencies: Record<string, string>
-      missingDependencies: Array<string>
-    },
-  ]
+  /**
+   * A run finished. Carries nothing: the file list is on `kubb:build:end`, the count is on
+   * `kubb:generation:summary`, contents are fetched with `studio:files`, and the dependency
+   * metadata travels with the `agent:snapshot` reply that needs it.
+   */
+  'kubb:generation:end': []
```
stijnvanhulle: Can we still have the same options as generation:start

stijnvanhulle:
**Comment ratio** is still red (21.6%, was 25.2%) after trimming `protocol/index.ts` and `StudioSession.ts`'s JSDoc in `d135c24`. I dug into why `packages/studio/src/ws.ts` alone still reports +3 comment lines despite that edit *removing* more comment text than it added, and found a reproducible bug in `tokei` 12.1.2 (the version this check pins), not a real excess of comments.

Isolating the one hunk that merges `ws.ts`'s `skipStorage`/`onGenerationEnd` doc comments into one block: with the file's *other* nine `/**` blocks untouched, going from 2 blocks to 1 in that spot (an odd total) makes tokei misattribute ~4 unrelated lines elsewhere in the file from code to comments. Going to 0 blocks there (even total) reports correctly again. The miscount is driven entirely by whether the file's total `/**...*/` block count is odd or even — verified by swapping in completely different comment text/length at that same spot and getting identical (wrong) numbers every time, and by removing the block entirely and watching the count return to normal.

I'm not going to chase this further by artificially splitting one field's doc comment into two disconnected blocks (or deleting real documentation) just to flip the parity back to even — that would make the code worse to satisfy a counting bug in the tool, not a real reduction in comment density. The actual comment-to-code ratio in this diff has genuinely dropped (54→44 added comment lines against unchanged added code), and CodeRabbit's two actionable findings (changeset bump level, mixed-version accuracy) are both addressed.

Leaving this check red with this explanation rather than contorting the diff further. Happy to help fix the check's config (e.g. an exclude or a version bump for `comment-ratio`) if a maintainer wants that route instead.

---

